PR #3, opened 2022-09-14
Add go.mod and go.sum for the certifier service

rgerganov:
The certifier service cannot be built unless GO111MODULE=off is set. This patch fixes this by adding go.mod and go.sum files.


rgerganov:
The module name also needs to be updated from `github.com/jlmucb/crypto/v2/certifier-framework-for-confidential-computing/certifier_service` to `github.com/vmware-research/certifier-framework-for-confidential-computing/certifier_service` to match the new repository. It will require changing the import path in the go sources. I can submit a new PR for this or change this one.

PR #18, opened 2023-02-01
Platfom verify

jlmucb:
This is the platform property support


src/certifier_proofs.cc
```diff
@@ -639,6 +639,21 @@ bool get_debug_property(const sev_attestation_message& sev_att, property* prop)
...
 
+bool get_tcb_version_property(const sev_attestation_message& sev_att, property* prop) {
+  int value = 0;
+
+  attestation_report* r= (attestation_report*) sev_att.reported_attestation().data();
+  value = (int)(((r->policy)>>8)&0xff);
+  string str_name("tcb-version");
+  string str_equal("=");
+  string str_type("int");
+  value = (((int)r->platform_version.boot_loader) << 24) |
```
yelvmw: I would just use the "r->raw" field here.

PR #29, opened 2023-03-14
Gramine/SGX DCAP remote verification - initial version

asmprogrammer5:


asmprogrammer5:
Not needed for now, since verification is done within Certifier.

Issue #33, opened 2023-03-30
Attestation Report is out of date.


larrydewey:
This attestation can be updated to match the current specification. Pulling an attestation report based off of this with the latest firmware will cause fields to be misaligned. 

https://github.com/vmware-research/certifier-framework-for-confidential-computing/blob/e78b65d6153be9251e07ef9f7a1cf5c2b91d9900/src/sev-snp/attestation.h#L54

larrydewey:
I will create a PR from my fork

yelvmw:
I did notice some updates. We pulled the report definition around 2021. Thanks for the reminder!

jlmucb:
Thanks, Larry.  We'll fix all of these.

jlmucb:
updated

PR #38, opened 2023-04-05
snp: Updating Attestation Report

larrydewey:
Updated the Attestation Report Structure to be compliant with newer firmware versions.

Resolves: #33


jlmucb (approved):
We need to check this against the existing SEV tests but it looks right.

PR #43, opened 2023-04-25
Rewrite instructions.txt in md-file formatting.

gapisback:
Rewrite sample_apps/simple_app/instructions.txt instructions in md-file format, in new file instructions.md .

---
NOTE: To the reviewer. 

The final form of simple_app/instructions.md can be found in my [dev-branch, here](https://github.com/vmware-research/certifier-framework-for-confidential-computing/blob/agurajada/simple_app-Instr.md/sample_apps/simple_app/instructions.md).

The contents are identical to current version of instructions.txt but just re-formatted to use md-style formatting. The new style is more readable and the commands to be issued stand-out on their own.

If this new md-format is acceptable, I will delete instructions.txt as part of this change-set.

Over time, the instructions.txt in other sample programs can also be converted to the new style as I go through each sample program one at a time.

### DIFF
--- a/sample_apps/simple_app/instructions.md
+++ b/sample_apps/simple_app/instructions.md
@@ -1,0 +1,468 @@
+# Simple App - Instructions 
+
+This document gives detailed instructions for building and running the sample application and generating the policy for the Certifier Service using the policy utilities.
+
+This simple_app sample program but provides an example of initializing
+and provisioning the Certifier Service with utility generated keys, measurements and policy.
+
+The sample program will still need to construct the statement "The attestation-key says the
+enclave-key speaks-for the program".  This is the attestation.
+
+Except for the ancillary files attest_key_file.bin, example_app.measurement and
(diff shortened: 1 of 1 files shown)


sample_apps/simple_app/instructions.md
```diff
@@ -0,0 +1,468 @@
...
+
+
+## Step 3: Generate the policy key and self-signed cert
+
+```shell
+$ cd $EXAMPLE_DIR/provisioning
+
+$ $CERTIFIER_PROTOTYPE/utilities/cert_utility.exe        \
+      --operation=generate-policy-key-and-test-keys      \
+      --policy_key_output_file=policy_key_file.bin       \
```
gapisback, line 57: I specifically preferred this style of listing each `--<option>` one on each line, so that each option stands out on its own.

I found that in the older format (in instructions.txt) with multiple `--options` on one line, it gets difficult to see what the invocation for each command actually is doing.

PR #44, opened 2023-04-25
Fix 'clean' recipe  to avoid getting noisy messasges. Update .gitignore rules

gapisback:
Fix diff Makefile's 'clean' recipe to use "rm -rf" on *.o's so that we don't get noisy messages reporting that *.o is not found. These occur when someone attempts 'clean' on a new branch, before making any executables.

Add more patterns to .gitignore rules.

----

**NOTE**: To reviewers: This change-set is mostly a cleanup-type changes to reduce noisy messages when running `clean` in a new code branch. I ran into these while going thru diff steps / instructions for sample programs.

As part of the build(s), we seem to produce many artifacts in the code line itself. So this commit enhances existing rules in `.gitignore` to exclude some of the generated files. 

Please see my annotations to `.gitignore` before we finalize this change. We don't want to accidentally ignore a file that should be tracked under git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,14 @@
 *.o
 *.swp
 *.swo
+*.pb.h
+*.pb.cc
+*.pb.go
+*.so
+*.cmd
+*.exe
+*.bin
+app1_data/
+app2_data/
+simpleserver
+provisioning/
--- a/application_service/app_service.mak
+++ b/application_service/app_service.mak
@@ -52,9 +52,9 @@ all:	app_service.exe hello_world.exe send_request.exe test_user.exe
 
 clean:
 	@echo "removing object files"
-	rm $(O)/*.o
+	rm -rf $(O)/*.o
 	@echo "removing executable file"
-	rm $(EXE_DIR)/app_service.exe
+	rm -rf $(EXE_DIR)/app_service.exe
 
 hello_world.exe: hello_world.cc
 	@echo "hello_world.cc"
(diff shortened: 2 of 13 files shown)


.gitignore
```diff
@@ -2,3 +2,14 @@
...
+*.pb.cc
+*.pb.go
+*.so
+*.cmd
+*.exe
+*.bin
+app1_data/
+app2_data/
+simpleserver
+provisioning/
```
gapisback, line 15: The above look like valid r.e's to exclude some generated files. With these rules, now we are reduced to the following shortened output from `git status`:

```
sdb-fdb-build:[96] $ git status
On branch agurajada/fix-mak-files-rm
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	sample_apps/simple_app/policy_key.cc
	sev-snp-simulator/Module.symvers
	sev-snp-simulator/modules.order
	sev-snp-simulator/sevnull.ko
	sev-snp-simulator/sevnull.mod

nothing added to commit but untracked files present (use "git add" to track)
```
I don't know if I should add specific rules for these files, or let it go. I suspect these are all also generated files, so could be git-ignore'ed.

Please advise.

application_service/app_service.mak
```diff
@@ -52,9 +52,9 @@ all:	app_service.exe hello_world.exe send_request.exe test_user.exe
 
 clean:
 	@echo "removing object files"
-	rm $(O)/*.o
+	rm -rf $(O)/*.o
 	@echo "removing executable file"
-	rm $(EXE_DIR)/app_service.exe
+	rm -rf $(EXE_DIR)/app_service.exe
```
gapisback, line 57: As an example, without these changes, `make clean` would complain and fail with non-zero `$rc` as follows:

```
sdb-fdb-build:[32] $ make -f example_app.mak clean
removing object files
rm ./*.o
rm: cannot remove './*.o': No such file or directory
make: *** [example_app.mak:50: clean] Error 1

sdb-fdb-build:[33] $ echo $?
2
```

In future, if we ever build a CI-system, this non-zero `$rc` from `make clean` will needlessly cause automation scripts to fail.

So, this kind of rework is better and also improves usability by not printing noisy messages.

PR #45, opened 2023-04-25
Relocate copyright notice to be at top of the file. Other minor edits.

gapisback:
This commit relocates the Copyright notice to be at the top of the file so it's consistently placed at the very top in all files. In a few files, user #include's are re-arranged so that they follow system .h #includes. 

There are no other code-/logic-changes with this commit.

----
**NOTE: To reviewers**: This change-set is mostly of a cleanup-nature. However, lots of source files are affected.

In principle, by simply rearranging some user-#include's to appear -after- system #include's `.h` files should not regress compilation. We don't quite have a full build-and-test CI. So, what's the way to ensure that these source file edits do not regress compilation? 

I have left this PR in 'draft' - stage just so that we can talk about this change, and to figure out how to test these diffs to ensure that everything still builds cleanly.

### DIFF
--- a/include/cc_helpers.h
+++ b/include/cc_helpers.h
@@ -1,19 +1,3 @@
-#include "support.h"
-#include "certifier.h"
-#include "simulated_enclave.h"
-
-#include <sys/socket.h>
-#include <arpa/inet.h>
-#include <netinet/in.h>
-#include <netdb.h>
-#include <openssl/ssl.h>
-#include <openssl/rsa.h>
-#include <openssl/x509.h>
-#include <openssl/evp.h>
-#include <openssl/rand.h>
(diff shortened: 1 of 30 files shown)


include/cc_helpers.h
```diff
@@ -28,6 +12,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <sys/socket.h>
```
gapisback, line 15: This is just a 'relocation' diff, only to ensure that the copyright notice consistently appears at the top of each file, wherever it appears.

jlmucb:
looks good

jlmucb:
looks fine

jlmucb (approved):

PR #47, opened 2023-04-26
Improve error handling and return of failure $rc for simple_app

gapisback:
This commit tightens some error-handling and reporting failures in helper methods. These were discovered while trying to setup simple_app in a new env.

- server_dispatch() now returns 'bool'. Couple of usages in the code-flow of simple_app are updated to deal with this failure.

- Cleaned-up and improved some status / progress printf() messages
- Replaced, in some places, inline name of function with __func__.
-  Additionally, some warnings in support.cc are fixed. To improve overall code quality, I am introducing "-Werror" in all Makefiles. This way, in future, if compiler warnings creep-in, that will cause build failures. This will force us to keep the code free of warnings, too.

----
**NOTE: To reviewer(s)**: 

While setting up `sample_app/simple_app` in my Nimbus-VM environment, I ran into some errors.
I was also working on packaging the entire set of instructions into a simple `run_example.sh` bash script (Not included in this change-set). When one or more of the steps in the instructions would fail, some of the helper methods were not correcting returning failure error code to the caller. `server_dispatch()` is one such example.

When a specific step fails to execute correctly but the failure is not returned to the shell, the bash script continues to run. It is supposed to abort at the point of the 1st failure.

Under this commit, I am trying to do local improvements to the `$rc` reporting in code-paths that encountered some errors while attempting to setup `sample_app/simple_app` .

If the nature of this change is acceptable, I will try to apply similar constructs to other sample programs, to improve the error-handling overall with the example programs.

### DIFF
--- a/include/cc_helpers.h
+++ b/include/cc_helpers.h
@@ -173,9 +173,9 @@ public:
   bool get_peer_id(string* out);
 };
 
-void server_dispatch(const string& host_name, int port,
-      string& asn1_root_cert, key_message& private_key,
-      const string& private_key_cert, void (*)(secure_authenticated_channel&));
+bool server_dispatch(const string& host_name, int port,
+                     string& asn1_root_cert, key_message& private_key,
+                     const string& private_key_cert, void (*)(secure_authenticated_channel&));
 
 #endif
 
--- a/sample_apps/simple_app/example_app.cc
+++ b/sample_apps/simple_app/example_app.cc
@@ -104,16 +104,6 @@ void server_application(secure_authenticated_channel& channel) {
   channel.write(strlen(msg), (byte*)msg);
 }
 
-bool run_me_as_server(const string& host_name, int port,
-      string& asn1_policy_cert, key_message& private_key,
-      string& private_key_cert) {
-
-  printf("running as server\n");
-  server_dispatch(host_name, port, asn1_policy_cert, private_key,
-      private_key_cert, server_application);
-  return true;
-}
-
 int main(int an, char** av) {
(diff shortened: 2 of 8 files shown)


include/cc_helpers.h
```diff
@@ -173,9 +173,9 @@ class secure_authenticated_channel {
   bool get_peer_id(string* out);
 };
 
-void server_dispatch(const string& host_name, int port,
-      string& asn1_root_cert, key_message& private_key,
-      const string& private_key_cert, void (*)(secure_authenticated_channel&));
+bool server_dispatch(const string& host_name, int port,
```
gapisback, line 176: Changed to return failure status.

sample_apps/simple_app/example_app.cc
```diff
@@ -104,16 +104,6 @@ void server_application(secure_authenticated_channel& channel) {
   channel.write(strlen(msg), (byte*)msg);
 }
 
-bool run_me_as_server(const string& host_name, int port,
```
gapisback, line 107: Discovered that this is an unused function. So, deleted it. (Let me know if you see any further need for this such that we should retain it.)

sample_apps/simple_app/example_app.cc
```diff
@@ -196,7 +186,7 @@ int main(int an, char** av) {
       goto done;
     }
 
-    printf("running as client\n");
+    printf("Running App as client\n");
```
gapisback, line 189: Minor improvement in message.

jlmucb (approved):

PR #49, opened 2023-04-27
Add SetUsageMessage() and usage_str to set of programs in utilities/

gapisback:
Enhance few programs in utilities/ sub-dir to call gflags::SetUsageMessage(), so that basic description of the program can be seen via --help output. Add, where needed, usage_str(), to report basic usage information.

------
**NOTE: To reviewer(s)**: 

This is just a sample set of changes to add improved help / usage messages to a small collection of programs. I think, by consistently reporting a small one-liner explaining what the purpose of each program is, we can improve the on-boarding experience for new users.

The actual content of the message will need to be hand-crafted for each program. I do not have sufficient familiarity with each program to craft a message. So, the bulk / brunt of the review process will be for the code-author and me to sit down and hash-out a descriptive message for each program.

### DIFF
--- a/utilities/make_platform.cc
+++ b/utilities/make_platform.cc
@@ -24,11 +24,17 @@ DEFINE_string(properties_file, "",  "properties files");
 DEFINE_string(output, "",  "output file");
 
 int main(int an, char** av) {
+  string usage("Generate platform-specific serialization of certificate. RESOLVE: Fix msg");
+  gflags::SetUsageMessage(usage);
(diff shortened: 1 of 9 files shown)


jlmucb (approved):

utilities/make_property.cc
jlmucb, line 1: SHould be "Specify a platform property used in policy"

utilities/make_platform.cc
jlmucb, line 1: Should be "Construct platform characteristics for platform verification policy"

PR #52, opened 2023-04-28
Update crypto

jlmucb:
Corrected RNG, added gcm authenticated encryption and deprecated asylo


src/support.cc
```diff
@@ -528,135 +528,264 @@ bool digest_message(const char* alg, const byte* message, int message_len,
...
+  int len;
+  int plaintext_len;
+  int stream_len = in_len - 2 * blk_size;
+  int err = 0;
+
+  if(!(ctx = EVP_CIPHER_CTX_new())) {
+    return false;
+  }
+  if(!EVP_DecryptInit_ex(ctx, EVP_aes_256_gcm(), nullptr, nullptr, nullptr)) {
+    ret = false;
```
gapisback, line 717: Suggestion: In this new chunk of code, there is a repetition of `ret = false;`. Instead, please consider doing this:

- Delete L704 where `ret` is declared and init'ed to `true`. Move that to L716, just before where `ret` is referenced.
- Just before L716, declare & init `bool ret = false;` once.
- Remove all the other initializations of `ret = false;` and just do `goto done`.
- On L752, just before `done: ` label, add `ret = true;`
 
This will clean up the code and get the right behaviour you were intending with all these `goto`s.

src/support.cc
```diff
@@ -528,135 +528,264 @@ bool digest_message(const char* alg, const byte* message, int message_len,
...
+done:
+  if (ctx != nullptr)
+    EVP_CIPHER_CTX_free(ctx);
+  return ret;
+}
+
+bool authenticated_encrypt(const char* alg_name, byte* in, int in_len, byte *key,
+            byte *iv, byte *out, int* out_size) {
+
+  if (strcmp(alg_name, "aes-256-cbc-hmac-sha256") == 0) {
```
gapisback, line 762: As this is new code, I'm making this comment. You may choose to do nothing right now.

Qs. Is `alg_name` null-terminated?

We are advised to use `strncmp()` for safe-comparison and to avoid comparing run-away string lengths.

I see that existing code uses `strcmp()` liberally, so perhaps, we can shelf this idea for now. In a future rev, I'd be happy to work with Ye & you to introduce `STRCMP()` and `STRNCMP()` as boolean comparison macros and tighten up this code.

gapisback (approved):
Took a quick look at the diffs. I understood nothing about the logic of these encryption-related changes. Just viewed the structure of the changes ... but haven't dug deepinto the logic.

Most of my comments are suggestions for improving the code hygiene. Some of these we can take up in future revs, after discussing what is a good way to implement stuff like, e.g, string comparison, string constants etc.

PR #53, opened 2023-05-08
Add README.md to sample_apps/ with brief description of each app.

gapisback:
Provide high-level description of each sample-app and the platform it runs on.


gapisback:
HI, @jlmucb - @yelvmw and I have gone over the text and have updated it in an internal review. He is now fine with this form of the contents.

Can you take a quick look and we can finalize the wording. (We can do it tomorrow when you are in the office, may be.)

sample_apps/README.md
jlmucb, line 1: As we discussed, move all the asylo stuff to deprecated and remove the asylo part of this doc.

For application enclave, the description should be
The example currently runs under the simulated enclave and should run any VM based platform host. Of the VM based platforms, it has been tested under SEV-SNP.

We should include a standard SEV test.

jlmucb (approved):

PR #56, opened 2023-05-09
Add rm_non_git_files() that only does 'git ls-files' for now.

gapisback:
Remove dependency of simple_app/run_example.sh on existence of src/certifier.pb.cc

This commit cleanses some logic in simple_app/run_example.sh to clean-up all non-git-tracked artifacts (.o's, .exe's, .cc/.h files etc.) from the dev area where steps to build-and-run simple_app are executed.

Previously, (in CI), other steps would have run 'make -f certifier.mak' in src/ sub-dir, that produces the required src/certifier.pb.cc file.

Rework the rules in utilities/policy_utilities.mak to correctly compile the protobuf file (from src/) to generate *.pb.cc/*.pb.h in the required sub-dirs.
Fix shellcheck errors in shell script.

### DIFF
--- a/sample_apps/simple_app/run_example.sh
+++ b/sample_apps/simple_app/run_example.sh
@@ -29,7 +29,6 @@ Client_app_data="./app1_data/"
 Srvr_app_data="./app2_data/"
 
 SimpleServer_PID=0
-AppAsServer_PID=0
 Found_step=0        # While looking thru Steps[] for user-specified fn-name
 
 # ###########################################################################
@@ -67,18 +66,19 @@ function usage() {
 
    # Computed elapsed hours, mins, seconds from total elapsed seconds
    echo "Usage: $Me [--help | --list]"
-   echo "  To run the example program          : ./${Me}"
-   echo "  To setup the example program        : ./${Me} setup"
-   echo "  To run and test the example program : ./${Me} run_test"
-   echo "  To list the individual steps        : ./${Me} --list"
-   echo "  To run an individual step           : ./${Me} <step name>"
+   echo "  To setup and run the example program, end-to-end : ./${Me}"
+   echo "  To setup the example program                     : ./${Me} setup"
+   echo "  To run and test the example program              : ./${Me} run_test"
+   echo "  To list the individual steps                     : ./${Me} --list"
+   echo "  To run an individual step                        : ./${Me} <step name>"
 }
 
 # ###########################################################################
 # Dump the list of functions, one for each step, that user can invoke.
 # Ref: https://www.freecodecamp.org/news/bash-array-how-to-declare-an-array-of-strings-in-a-bash-script/
 # ###########################################################################
-Steps=("show_env"
+Steps=( "rm_non_git_files"
+        "show_env"
         "do_cleanup"
         "build_utilities"
         "gen_policy_and_self_signed_cert"
@@ -154,6 +154,26 @@ function run_cmd() {
    "$@"
 
    set +x
+}
+
+# ###########################################################################
+# Deep-clean the build-env to remove artifacts (e.g. generated files, binaries
+# etc.) that may have been produced by other steps. We run this to ensure
+# that this script will run successfully w/o any dependencies on executing
+# some prior steps.
+# ###########################################################################
+function rm_non_git_files() {
+    pushd "${CERT_PROTO}" > /dev/null 2>&1
+
+    echo "${Me}: Delete all files not tracked by git"
+    # shellcheck disable=SC2046
+    run_cmd rm -rf $(git ls-files . --exclude-standard --others)
+
+    echo "${Me}: Delete all files not tracked by git that are also ignored"
+    # shellcheck disable=SC2046
+    run_cmd rm -rf $(git ls-files . --exclude-standard --others --ignored)
+
+    popd > /dev/null 2>&1
 }
 
 # ###########################################################################
@@ -551,18 +571,25 @@ function run_app_as_client_make_trusted_request() {
 
 # ###########################################################################
 function show_env() {
-   echo "Environment variables, script globals:"
+   echo
+   echo "**** Environment variables, script globals: ****"
    env | grep -E "CERT_PROTO|EXAMPLE_DIR"
 
-   local numCPUs=`grep "^processor" /proc/cpuinfo | wc -l`
-   local cpuModel=`grep "model name" /proc/cpuinfo | head -1 | cut -f2 -d':'`
-   local cpuVendor=`grep "vendor_id" /proc/cpuinfo | head -1 | cut -f2 -d':'`
-   local totalMemGB=`free -g | grep "^Mem:" | awk '{print $2}'`
+   local numCPUs=0
+   local cpuModel=0
+   local cpuVendor=0
+   local totalMemGB=0
+
+   numCPUs=$(grep -c "^processor" /proc/cpuinfo)
+   cpuModel=$(grep "model name" /proc/cpuinfo | head -1 | cut -f2 -d':')
+   cpuVendor=$(grep "vendor_id" /proc/cpuinfo | head -1 | cut -f2 -d':')
+   totalMemGB=$(free -g | grep "^Mem:" | awk '{print $2}')
 
     echo
     uname -a
-    echo "${cpuVendor}, ${numCPUs} CPUs, ${totalMemGB} GB, ${cpuModel}"
-    ping -4 -c 1 `uname -n` | head -2
+    echo
+    echo "${Me}: ${cpuVendor}, ${numCPUs} CPUs, ${totalMemGB} GB, ${cpuModel}"
+    ping -4 -c 1 "$(uname -n | head -2)"
 
     echo
     lsb_release -a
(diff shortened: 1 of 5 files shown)


sample_apps/simple_app/run_example.sh
```diff
@@ -29,7 +29,6 @@ Client_app_data="./app1_data/"
 Srvr_app_data="./app2_data/"
 
 SimpleServer_PID=0
-AppAsServer_PID=0
```
gapisback, line 32: Deleted because `shellcheck` complains about unused variable.

sample_apps/simple_app/run_example.sh
```diff
@@ -156,6 +156,26 @@ function run_cmd() {
    set +x
 }
 
+# ###########################################################################
+# Deep-clean the build-env to remove artifacts (e.g. generated files, binaries
+# etc.) that may have been produced by other steps. We run this to ensure
+# that this script will run successfully w/o any dependencies on executing
+# some prior steps.
+# ###########################################################################
+function rm_non_git_files() {
```
gapisback, line 165: This new function will always run, to delete non-Git-owned objects.

This helps test the makefile changes in an env where nothing generated previously exists.

sample_apps/simple_app/run_example.sh
```diff
@@ -156,6 +156,26 @@ function run_cmd() {
...
+    pushd "${CERT_PROTO}" > /dev/null 2>&1
+
+    echo "${Me}: Delete all files not tracked by git"
+    # shellcheck disable=SC2046
+    run_cmd rm -rf $(git ls-files . --exclude-standard --others)
+
+    echo "${Me}: Delete all files not tracked by git that are also ignored"
+    # shellcheck disable=SC2046
+    run_cmd rm -rf $(git ls-files . --exclude-standard --others --ignored)
```
gapisback, line 174: NOTE: I have manually tested these two `git ls-files` commands, and they _seem_ to work well-enough for our current usages.

I am no Git-master ,... so am relying on CI-clean test runs to validate that this change is acceptable.

sample_apps/simple_app/run_example.sh
```diff
@@ -554,15 +574,21 @@ function show_env() {
...
-   local totalMemGB=`free -g | grep "^Mem:" | awk '{print $2}'`
+   local numCPUs=0
+   local cpuModel=0
+   local cpuVendor=0
+   local totalMemGB=0
+
+   numCPUs=$(grep -c "^processor" /proc/cpuinfo)
+   cpuModel=$(grep "model name" /proc/cpuinfo | head -1 | cut -f2 -d':')
+   cpuVendor=$(grep "vendor_id" /proc/cpuinfo | head -1 | cut -f2 -d':')
+   totalMemGB=$(free -g | grep "^Mem:" | awk '{print $2}')
```
gapisback, line 585: This whole chunk fixes errors reported by `shellcheck`.

utilities/policy_utilities.mak
```diff
@@ -28,8 +28,11 @@ TARGET_MACHINE_TYPE= x64
 endif
 
 S= $(SRC_DIR)
-CERT_SRC=$(CERTIFIER_PROTOTYPE_DIR)/src
 O= $(OBJ_DIR)
+I= $(INC_DIR)
```
gapisback, line 32: Previously `I` and `US` (one line below) were not defined, but were being used in the make commands.

(Surprisingly, things somehow worked ok ...) 

I defined these here based on how they were being defined in the companion `cert_utility.mak` file.

utilities/policy_utilities.mak
```diff
@@ -116,9 +119,9 @@ $(O)/key_utility.o: $(US)/key_utility.cc $(INC_DIR)/support.h $(INC_DIR)/certifi
 	@echo "compiling key_utility.cc"
 	$(CC) $(CFLAGS) -c -o $(O)/key_utility.o $(US)/key_utility.cc
 
-#$(US)/certifier.pb.cc $(I)/certifier.pb.h: $(CERT_SRC)/certifier.proto
-#	$(PROTO) -I$(S) --cpp_out=$(US) $(CERT_SRC)/certifier.proto
-#	mv certifier.pb.h $(I)
+$(CERT_SRC)/certifier.pb.cc $(I)/certifier.pb.h: $(CERT_SRC)/certifier.proto
+	$(PROTO) -I$(S) --proto_path=$(CERT_SRC) --cpp_out=$(CERT_SRC) $(CERT_SRC)/certifier.proto
```
gapisback, line 123: The real-fix seems to be the use of `--proto_path=$(CERT_SRC)`. I found this in `man git`.

The change on L122 cleans-up some dependency listing ... which seem ok to me.

Again, I am no Make-expert ... so another pair of eyes during review will be good!
All this mess seems to work fine in manual & CI testing.

utilities/policy_utilities.mak
```diff
@@ -116,9 +119,9 @@ $(O)/key_utility.o: $(US)/key_utility.cc $(INC_DIR)/support.h $(INC_DIR)/certifi
...
+	mv $(CERT_SRC)/certifier.pb.h $(I)
```
gapisback, line 124: This _seems_ correct; i.e. the generated `.h` file will be in `certifier/src` sub-dir.

sample_apps/simple_app/run_example.sh
```diff
@@ -80,6 +79,7 @@ function usage() {
 # ###########################################################################
 Steps=("show_env"
         "do_cleanup"
+        "rm_non_git_files"
```
gapisback, line 82: We should consider an option here:

Currently, with this fix, we will delete all non-Git files each time you run this script.

John wanted some flexibility to retain outputs of previously run steps, so this script's execution can be optimized.

That can be 'controlled' by location of this new step `rm_non_git_files` in this area.

I can move it to the top. And do the following:

- If you run `run_example.sh` by itself, it will start from the top, so `rm_non_git_files` will be invoked.
- If you run `run_example.sh setup`, that's an indication that we may want to retain existing artifacts. So execution will begin from 2nd step, i.e. `show_env`. 

This way, one gets a bit more flexibility of how this script can be used.

In CI, we are running `./run_example.sh` without any arguments, so you will get the rm-non-Git-objects behaviour.

Let me know what you prefer.

yelvmw: I think this makes sense. I'll leave it to John.

utilities/policy_utilities.mak
yelvmw: Thanks for fixing this. We might want to do something similar in cert_utility.mak too so that we never duplicate the protobuf generation.

gapisback:
@yelvmw -- About your concern / comment that we are re-running `protoc` twice in both `utilities`/*.mak` files ... I tested this manually after cleaning up all non-Git stuff.

The rule from `cert_utility.mak` populates the output in `US`; i.e. current dir:

```
/home/agurajada/Code/certifier-framework-for-confidential-computing/utilities

sdb-fdb-build:[33] $ ll *.pb*
-rw-rw-r-- 1 agurajada   610396 May  9 15:50 certifier.pb.cc
-rw-rw-r-- 1 agurajada 12648960 May  9 15:50 certifier.pb.o
```

The rule from `policy_utilities.mak` outputs the file in `--cpp_out=$(CERT_SRC)`; i.e., `certifier/src/` sub-dir.

Here is the confirmation:

```
sdb-fdb-build:[33] $ ll ../src/*pb*
-rw-rw-r-- 1 agurajada 610396 May  9 15:52 ../src/certifier.pb.cc
```

I am not very bothered by these two invocations of `protoc` as it's a quick step, and these build-time commands are reasonably fast.

utilities/policy_utilities.mak
gapisback: I really don't know what this duplication is all about.

Both `protoc` invocations are emitting outputs to diff sub-dirs, to satisfy other build rules.

I'd like to not touch the other .mak file at this time. Let this go in .. and we can re-check with John once he gets to use this and test it out in his own env.

yelvmw (approved):
LGTM!

gapisback:
Hi, @jlmucb -- @yelvmw has reviewed this and he'd like you to take a look, too, as the changes involve invocation of `protoc`.

He feels that we may want to optimize (or avoid one of the two calls to) `protoc` that occur in `utilities/cert_utility.mak` and the other one (which I reworked) in `utilities/policy_utilities.mak`.

Also, I have a suggestion for your consideration in the `run_example.sh` on L82, to decide where to place this newly added `rm_non_git_files` step. See my comments in that file. We could add a 'feature' that goes towards something you wanted to be able to do when exercising this shell script.

sample_apps/simple_app/run_example.sh
gapisback: @jlmucb - To expedite decision-making, I decided to go with this new proposed change.

`run_example.sh` w/o any args will always invoke the newly added `rm_non_git_files()` function, so we will now exercise this script from  a clean-cloned repo build-env.

I also extended `build.yml` to exercise this combination of `setup`, followed by one or more `run_test` commands.

Once this works in CI, we can have more confidence that this interface will stay stable and useful / usable by customers.

jlmucb (approved):
No additional changes to this from me.

jlmucb (approved):
No additional changes to this from me.

PR #59, opened 2023-05-11
Allow specification of authenticated encryption algorithm for Protect

jlmucb:
This small change lets us specify the authenticated encryption algorithm for Protect_Blob.  The default is "aes-256-cbc-hmac-sha256" but you can also specify "aes-256-gcm"."  This  keeps us in conformance with CNSA 2.0.


yelvmw (approved):
LGTM!

PR #61, opened 2023-05-12
Export all RSA key properties

rgerganov:
There are usecases when we need to export the RSA private key from the policy store. One example is running an HTTPS endpoint with the certificate obtained from the Certifier service. In such cases we need to save all key properties (n, e, d, p, q, dmp1, dmq1, iqmp) in order to reconstruct the key outside of the policy store.


jlmucb (approved):
this looks OK

jlmucb (approved):
this looks OK

Issue #64, opened 2023-05-12
Prototype: Move simple_app/run_example.sh to sample_apps/ dir and rework to drive diff build-and-test of diff apps.


gapisback:
In a previous commit, `simple_app/run_example.sh ` was developed and integrated to /main. It works just for the `simple_app`.

This script was copy-cloned and extended to run for `simple_app_under_oe/`. With that work, the diffs between these two scripts is very minimal. Investigation of `instructions.txt` of other sample apps shows that _most_ of the steps are common and similar for all sample programs.

So, under this issue:

1. Relocate `simple_app/run_example.sh` to one-level up; i.e. `sample_apps/` dir.
2. Fix the arg-parsing, `--help` etc. to work with a sample-app program name; i.e. support a usage something on the lines of: 

   $ run_example.sh simple_app
   $ run_example.sh --help simple_app
   
3. If this works and passes CI, integrate this change to /main. Then this script can be extended to support `simple_app_under_oe/`, for which the dev-version of this script already exists in dev-branch `agurajada/simple_app_under_oe-Fix-Bugs`

PR #65, opened 2023-05-15
Move run_example.sh & cleanup.sh from simple_app/ to sample_apps/

gapisback:
This commit pre-stages the location of two scripts that were developed for setting up and running the simple_app/ . The idea is to parametrize run_example.sh so that it can be used to build-and-test more than a few sample apps, starting from simple_app/. This extension will be done in follow-on PRs.


gapisback:
Attn: @jlmucb - This PR simply re-locates two scripts up one level, so that future changes to address issue #64 can be implemented, with few diffs in the change-set.

Issue #67, opened 2023-05-15
Extend sample_apps/run_example.sh to build and run simple_app_under_oe program


gapisback:
Upcoming PR #66 extends `run_example.sh` to run `simple_app`. Under draft PR #62 a version of this script has been developed to run `simple_app_under_oe`. We need to consolidate both scripts so that the same `sample_apps/run_example.sh` script can be used to run this new app as well.

PR #70, opened 2023-05-15
(#67) Extend common run_example.sh for simple_app_under_oe/ program.

gapisback:
Rework --help / --list, arg interfaces to support OE

    This commit simply adds new step-fns to support OE. And updates
    -h/--help, --list options. Nothing is executed, per se.

    New interfaces are:

     ./run_example.sh [-h | --help | --list] <sample-app-name> [ setup | run_test ]

    Add fixes to get simple_app running cleanly through this script on SGX platform.
    Extend test cases in build.yml to exercise these new interfaces.

    Following now works on SGX and also on Nimbus-VM:

     - Add $HOME/go/bin to PATH to find
     - ./run_example.sh simple_app


sample_apps/README.md
```diff
@@ -61,12 +61,11 @@ You can use this script to Build and run example programs.
 
 **Usage**:
 ```shell
-run_example.sh [-h | --help] < sample-app-name > [ setup | run_test | --list ]
+run_example.sh [-h | --help | --list] <sample-app-name> [ setup | run_test ]
```
gapisback, line 64: Made the interface a bit more consistent with `--help` and `--list` coming up-ahead.

sample_apps/run_example.sh
```diff
@@ -10,19 +10,31 @@ Me=$(basename "$0")
...
 Cleanup="$(pwd)/cleanup.sh"
 popd > /dev/null 2>&1
 
 MeMsg="Build and run example program"
 
 # Script global symbols to app-data dirs
-Client_app_data="./app1_data/"
-Srvr_app_data="./app2_data/"
+Client_app_data="app1_data"
+Srvr_app_data="app2_data"
```
gapisback, line 24: Sanitized: Now we have only one instance of these `app1_data` and `app2_data` in the scripts. 

Everywhere else, we now use the symbols.

sample_apps/run_example.sh
```diff
@@ -10,19 +10,31 @@ Me=$(basename "$0")
...
 Valid_app=0         # While looking thru SampleApps[] for supported app
 Found_step=0        # While looking thru Steps[] for user-specified fn-name
 
+# build_utilites needs access to OpenSSL libraries, that may live on a diff
+# dir location, depending on how OpenSSL was installed on diff machines.
+# User can over-ride our default by specifying LOCAL_LIB env-var
+Local_lib_path=${LOCAL_LIB:-/usr/local/lib}
+
+# Sub-tools used in this script
+Jq=jq    # Needed for OE app; will be established later on.
```
gapisback, line 36: Need for automating edit of json file to replace "stuff" in it.

sample_apps/run_example.sh
```diff
@@ -94,11 +108,22 @@ function usage() {
    echo "  To run an individual step of the ${SampleAppName}         : ./${Me} ${SampleAppName} <step name>"
 }
 
+# ---- Open-Enclave app-specific help/usage output
+function usage_OE() {
```
gapisback, line 112: best understood by running: `./run_example.sh --help simple_app_under_oe`

sample_apps/run_example.sh
```diff
@@ -94,11 +108,22 @@ function usage() {
...
+   echo "  To setup the example program        : ./${Me} simple_app_under_oe setup_with_generate_policy"
+   echo "  To run and test the example program : ./${Me} run_test"
+}
+
 # ###########################################################################
 # Dump the list of functions, one for each step, that user can invoke.
 # Ref: https://www.freecodecamp.org/news/bash-array-how-to-declare-an-array-of-strings-in-a-bash-script/
 # ###########################################################################
 Steps=( "rm_non_git_files"
+
```
gapisback, line 125: Each application now has its own set of "Steps"; i.e. a list of functions implementing one part of the procedure.

sample_apps/run_example.sh
```diff
@@ -107,27 +132,95 @@ Steps=( "rm_non_git_files"
...
+        # 'run_test' argument starts from here:
         "start_certifier_service"
         "run_app_as_server_talk_to_Cert_Service"
         "run_app_as_client_talk_to_Cert_Service"
         "run_app_as_server_offers_trusted_service"
         "run_app_as_client_make_trusted_request"
 )
 
+Steps_OE=( "rm_non_git_files"
```
gapisback, line 157: "Steps" for `simple_app_under_oe`, which shares many steps with `simple_app`.

sample_apps/run_example.sh
```diff
@@ -241,7 +352,28 @@ function emded_policy_in_example_app() {
 }
 
 # ###########################################################################
+# As this function is executed through a table of function names,
+# caller cannot pass-in the sample-app-name. So, fall back to using
+# a global variable.
 function compile_app() {
+    case "${SampleAppName}" in
```
gapisback, line 359: You will see this pattern in many such functions.

The entry-point. step function, `compile_app()`, farms off control to application-specific step function based on the sample-app-name established at the beginning of the script's execution.

sample_apps/run_example.sh
```diff
@@ -409,6 +679,88 @@ function print_signed_claim() {
    popd > /dev/null 2>&1
 }
 
+# ###########################################################################
+# Step 7 in instructions.md: Manual policy generation
```
gapisback, line 683: Step 7 here, and Step 8 (on L706) below refer to [contributions.md](https://github.com/vmware-research/certifier-framework-for-confidential-computing/blob/agurajada/67-run_example_for-oe-app/sample_apps/simple_app_under_oe/instructions.md). 

I need some help in better phrasing the section titles for these two steps in that md-file. Currently they read:

- Step 7: Manual policy generation
- Step 8: Use Policy Generator

I don't think these phrases convey the work that each type of this step is doing.

What is the difference between "Manual" and "Use Policy" ?
Can we come up with better section titles? I can then rework the comments / names of these functions in the shell script accordingly.

sample_apps/run_example.sh
```diff
@@ -409,6 +679,88 @@ function print_signed_claim() {
...
+    set -x
+    mrenclave=$(make dump_mrenclave | grep "^mrenclave" | cut -f2 -d'=')
+    set +x
+
+    # Update json file using 'jq' tool, to replace value of the
+    # 'measurement' key with generic term 'mrenclave'. Then do a text
+    # replacement of this with the mrenclave grabbed above.
+    set -x
+    Jq=$(command -v jq)
+    ${Jq} '.measurements = [ "mrenclave" ]' "${policy_json_file}"    \
```
gapisback, line 741: This introduces a dependency on external tool `jq` to do this in-place edit of the json file.

I found no other more reliable way.

I have doc'ed this dependency in the md-file [here, in the README.md](https://github.com/vmware-research/certifier-framework-for-confidential-computing/tree/agurajada/67-run_example_for-oe-app/sample_apps/simple_app_under_oe).

---
Setup the simple app by editing the policy JSON file and use policy generator: For this interface, you will need to install the jq utility for editing the JSON file. $ sudo apt-get install -y jq

sample_apps/run_example.sh
```diff
@@ -426,10 +778,13 @@ function build_simple_server() {
 
     # Compile the oelib for OE host verification
     # Likely, user does not have OE SDK installed or does not want to enable OE:
-    # This should produce a go file for the certifier protobufs
+    # This should produce a Go file for the certifier protobufs
     # called certifier.pb.go in certprotos.
+    make_arg=""
+    if [ "${SampleAppName}" = "simple_app" ]; then make_arg=dummy; fi
```
gapisback, line 784: We will do real `make` for OE-simple-app but will do `make dummy` for `simple_app`.

sample_apps/simple_app_under_oe/host/host.cc
```diff
@@ -140,15 +144,15 @@ int main(int argc, const char* argv[])
...
     }
 
-
 exit:
     // Clean up the enclave if we created one
     if (enclave)
         oe_terminate_enclave(enclave);
 
-    return ret;
+    return 0;
```
gapisback, line 157: Change `ret` to always return 0; otherwise, even due to benign failures this program will return non-zero $rc, causing shell script to exit early.

sample_apps/run_example.sh
```diff
@@ -107,27 +132,95 @@ Steps=( "rm_non_git_files"
...
         "get_measurement_of_trusted_app"
         "author_policy"
-        "construct_policyKey_platform_is_trusted"
-        "construct_policyKey_measurement_is_trusted"
-        "combine_policy_stmts"
-        "print_policy"
-        "construct_platform_key_attestation_stmt_sign_it"
-        "print_signed_claim"
+            # These sub-step fns are subsumed under author_policy()
+            "construct_policyKey_platform_is_trusted"
```
gapisback, line 136: These sub-steps are subsumed (and run by) `author_policy`.

It is difficult to indicate through this array that user has to only run `author_policy` or run this list of sub-steps.

The script itself takes care of this duplication.

We can try to document it in the instructions.md or README.md, but I feel this business is very ver low-level.

And will be tedious to explain everything.

Comments welcome ... on how-to handle this nesting issue.

sample_apps/run_example.sh
```diff
@@ -647,11 +1149,64 @@ function run_steps() {
...
+}
+
+# ###########################################################################
+# Do initial setup of the sample app test-case
+# As the sub-steps array has nested fn-calls, run a collection of steps
+# to avoid re-running sub-fns nested under author_policy()
+# ###########################################################################
+function setup_simple_app() {
+    run_steps "show_env" "get_measurement_of_trusted_app"
+    author_policy
```
gapisback, line 1176: Because `author_policy` already calls some sub-step fns, we do break-up the execution to separately execute `author_policy`. Then, continue -after- the list of sub-step fns it includes.

Same gymnastics are done in `setup_simple_app_for_oe()` on L1194 onwards.

sample_apps/run_example.sh
gapisback: Also see my comments in the diff for `setup_simple_app()` and `setup_simple_app_for_oe()` which explains how the execution handles these nested sub-step functions list.

sample_apps/run_example.sh
```diff
@@ -647,11 +1149,64 @@ function run_steps() {
...
+    if [ "${setup_arg}" = "" ]; then
+        author_policy_for_OE
+    else
+        ${setup_arg}
+    fi
+    run_steps "build_simple_server" "provision_app_service_files"
+}
+
+# ###########################################################################
+function setup_with_generate_policy() {
```
gapisback, line 1203: This name needs a re-think. This is implementing the alternative Step 8: Use Policy Generator.

What better name can we come up to self-document the actual implementation of editing this json file etc.

sample_apps/simple_app/instructions.md
```diff
@@ -23,69 +23,68 @@ $CERTIFIER_PROTOTYPE is the top level directory for the Certifier repository.
 It is helpful to have a shell variable for it, e.g., :
 
 ```shell
-$ export CERTIFIER_PROTOTYPE=~/src/github.com/jlmucb/crypto/v2/certifier-framework-for-confidential-computing
+export CERTIFIER_PROTOTYPE=~/src/github.com/jlmucb/crypto/v2/certifier-framework-for-confidential-computing
```
gapisback, line 26: You can skip reviewing this file.

I basically had 2nd thoughts on adding `$ ` prefix to each command.

When there are multiple commands shown, it's much easier to mouse-over and cut-and-paste all 3-4 commands in one go w/o the leading `$ `. 

So the diffs in this file are mainly to take out this leading `$ `

sample_apps/simple_app_under_oe/Makefile
```diff
@@ -23,6 +23,7 @@ cleanall: clean
 
 dump_mrenclave:
 	oesign dump --enclave-image=./enclave/enclave.signed
+	@echo "\e[1;31m Build measurement_init.exe with make -f cert_utility.mak measurement_init.exe in utilities directory\e[1;m"
 	@echo "\e[1;31m Invoke measurement_init.exe --mrenclave=xxx --out_file=binary_trusted_measurements_file.bin\e[1;m"
 	@echo "\e[1;31m Replace 'xxx' with the mrenclave string from the output above\e[1;m"
-	@echo "\e[1;31m Build measurement_init.exe with make -f cert_utility.mak measurement_init.exe in utilities directory\e[1;m"
+	tput init
```
gapisback, line 29: I had to add this to re-set the red-char font that the previous message leaves the terminal in.

Otherwise, in an interactive usage, the rest of the commands will be in red, which is distracting.

This fix works for interactive execution of this script. But if you redirect output to a file, then again, this `tput init` seems to "confuse" the redirection. 

I noticed that the execution sort-of 'halts' at this point.

I had to manually run `tput init` in that terminal for the background'ed execution to continue.

All this is messy ... and we need a cleaner solution for both requirements (of displaying the message in red font and then to reset the stty terminal after).

Any ideas?

sample_apps/simple_app_under_oe/oe_policy.json
```diff
@@ -1,7 +1,6 @@
 {
-  "policyKey" : "policy_key_file.bin",
-
-  "measurements" : [
-    "c7037df41e773056809af62acf78a107d8e43a0277fbc3431e5539bc8e1ee12a"
+  "policyKey": "policy_key_file.bin",
```
gapisback, line 2: Should this file be updated every time? I think this diff should be withdrawn, but need input from the reviewers.

sample_apps/simple_app_under_oe/test.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+# ##############################################################################
+# test.sh: Execute build-and-test of simple_app_under_oe with diff arguments
```
gapisback, line 3: I added this driver shell script so that once all these changes stabilize, one can simply run this manually on an sgx-enabled box, to exercise all interfaces in one single script.

jlmucb (approved):
LGTM

jlmucb (approved):
LGTM

jlmucb (approved):
LGTM

sample_apps/run_example.sh
yelvmw: Manually generate certifier policy or generate the same policy using the policy generator. Only one of these is required.

sample_apps/run_example.sh
yelvmw: "Policy Generator" is the name of the tool

sample_apps/simple_app_under_oe/Makefile
yelvmw: Have you tried `echo -e "\e[0mNormal Text"`?

yelvmw: Essentially "\e[0m". The reset VT100 sequence? I can never remember these. See: https://m.blog.naver.com/PostView.naver?isHttpsRedirect=true&blogId=heehow&logNo=221603147435

sample_apps/run_example.sh
gapisback: Reworked names of functions and tags to:

- author_policy_for_OE() -> manual_policy_generation_for_OE()
- generate_policy_for_OE() -> automated_policy_generation_for_OE()
- Tag `setup_with_generate_policy` -> `setup_with_auto_policy_generation_for_OE`

sample_apps/simple_app_under_oe/Makefile
gapisback: Adding an extra final echo line, like `@echo "\e[0;m"` seems to have resolved this stty color issue.

`./test.sh` now runs w/o hanging. Thanks for the tip!

sample_apps/run_example.sh
gapisback: Changed fn's name as we discussed to `setup_with_auto_policy_generation_for_OE()`

sample_apps/simple_app_under_oe/host/host.cc
gapisback: Done. `return rv` also seems to work reliably now in all test scenarios.

PR #76, opened 2023-05-18
Certifier user header refactoring

yelvmw:
Refactored Certifier headers. All Certifier user facing public interfaces are moved to certifier_framework.h and all the Certifier utilities are moved to certifier_utilities.h. No other headers should be used by user applications. As a first step, some internal headers now include the public headers to maintain consistency. These will be cleaned up in the followup patches.

### DIFF
--- a/application_service/app_service.cc
+++ b/application_service/app_service.cc
@@ -27,6 +27,9 @@
 #include <linux/memfd.h>
 #include <sys/mman.h>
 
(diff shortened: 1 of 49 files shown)


jlmucb (approved):

jlmucb (approved):
Namespace stuff looks good

jlmucb (approved):
Namespace stuff looks good

Issue #72, opened 2023-05-18
Compiler error seen in printf() format specifier while building C-Certifier code on macOSX


gapisback:
Reported by @jlmucb - These errors are seen on MacOSX

g++ -I../include -I/usr/local/opt/openssl@1.1/include/ -I../src/sev-snp/ -O3 -g -Wall -std=c++11 -Wno-unused-variable -D X64 -Wno-deprecated -Wno-deprecated-declarations -Werror -c -o ./support.o ../src/support.cc
../src/support.cc:386:19: error: format specifies type 'unsigned long' but the argument has type
      '::google::protobuf::uint64' (aka 'unsigned long long') [-Werror,-Wformat]
    printf("%lu", prop.int_value());
            ~~~   ^~~~~~~~~~~~~~~~
            %llu
../src/support.cc:2107:20: error: format specifies type 'unsigned long' but the argument has type
      '::google::protobuf::uint64' (aka 'unsigned long long') [-Werror,-Wformat]
    printf(" %lu", p.int_value());
             ~~~   ^~~~~~~~~~~~~

PR #78, opened 2023-05-19
(#72) Fix print fmt-specifier that causes cc failure on macOSX

gapisback:
As reported in issue #72, couple of files are failing to compile on macOSX. This commit fixes those failures by using long long print specifier, `PRIu64`.
 
 Things seem to work fine on macOSX for `$ make -f cert_utility.mak`.

### DIFF
--- a/src/support.cc
+++ b/src/support.cc
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <inttypes.h>
 #include <openssl/ssl.h>
 #include <openssl/rsa.h>
 #include <openssl/x509.h>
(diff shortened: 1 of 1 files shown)


src/support.cc
```diff
@@ -383,7 +384,7 @@ void print_property(const property& prop) {
     } else if (prop.comparator() == ">=") {
       printf(" >= ");
     }
-    printf("%lu", prop.int_value());
+    printf("%" PRIu64, prop.int_value());
```
gapisback, line 387: Minor fix to use the right print format specifier, which is encoded by `PRIu64`. This is defined on each platform appropriate to what the compiler system requires.

yelvmw:
Are we sure this is the only place needed?

gapisback:
@yelvmw -- received confirmation from @jlmucb that this fix is good to go. I'm going to push this to `/main` shortly.

---
----- Forwarded Message -----
From: John Manferdelli 
To: Aditya Gurajada 
Sent: Sunday, May 21, 2023 at 02:33:08 PM PDT
Subject: Re: About fix for issue # 72 - printf build failure on macOSX

I'm good with it.  Assigning it to me was a mistake.  Go ahead and merge

Issue #73, opened 2023-05-18
run_example.sh: Simplify code that calls app-specific sub-fn-steps using $app_name variable.


gapisback:
In a recent checkin under issue #67 , `run_example.sh` was reworked to introduce support for multiple apps.

A sample code fragment like below can be simplified in interesting ways:

```
 367 function compile_app() {
 368     case "${SampleAppName}" in
 369         "simple_app")
 370             compile_simple_app
 371             ;;
 372
 373         "simple_app_under_oe")
 374             compile_simple_app_under_oe ""
 375             ;;
 376
 377         *)
 378             echo "${Me}: Unknown sample app name '${app_name}'."
 379             return
 380             ;;
 381     esac
 382 }
```

This can be simplified as follows:

With a consistent sub-fn naming convention, we can directly call `compile_${SampleAppName}`. 
I have tested this during dev, but did not apply that construct everywhere this pattern of coding exists in this script.

This re-written calling sequence can be applied to the following functions:

- setup()
- compile_app()
- get_measurement_of_trusted_app()
- construct_policyKey_platform_is_trusted()
- run_app_as_server_talk_to_Cert_Service() - may be will need some name rework.
- run_app_as_client_talk_to_Cert_Service()
- run_app_as_server_offers_trusted_service()
- run_app_as_client_make_trusted_request()

---
Finally, this block where we check for an invalid app-name can be deleted from all these functions as that check is already being done at the top-level of the script's execution:

```
1140 function setup() {
1141     local app_name="$1"
[...]
1151         *)
1152             echo "${Me}: Unknown sample app name '${app_name}'."
1153             return
1154             ;;
1155     esac
1156 }
```

Both these changes will:
- make the overall script shorter 
- tighten the consistency of naming convention of sub-step functions with the name of the sample-app being tested
- simplify folding in support for other sample-apps.

PR #79, opened 2023-05-19
(#73) run_example.sh: Rework sub-step fns using $SampleAppName variable

gapisback:
This commit simplifies the calling sequence of app-specific step- functions in run_example.sh to use variable name, $SampleAppName, whenever possible. This enforces a naming consistency requirement that step-functions are named, e.g., setup_simple_app() and setup_simple_app_under_oe() etc. This change makes it easier to integrate support for other sample apps under this bash script.

-----
NOTE to the reviewer: This refactoring is essentially no-code-change change set. As long as CI runs and the newly added test cases to exercise `--dry-run simple_app_under_oe` succeed, I am comfortable that these script changes should work cleanly on SGX platform for OE-apps, too.

### DIFF
--- a/sample_apps/run_example.sh
+++ b/sample_apps/run_example.sh
@@ -407,20 +407,7 @@ function emded_policy_in_example_app() {
 # caller cannot pass-in the sample-app-name. So, fall back to using
 # a global variable.
 function compile_app() {
-    case "${SampleAppName}" in
-        "simple_app")
-            compile_simple_app
-            ;;
-
-        "simple_app_under_oe")
-            compile_simple_app_under_oe ""
-            ;;
-
-        *)
-            echo "${Me}: Unknown sample app name '${app_name}'."
-            return
-            ;;
-    esac
+    compile_"${SampleAppName}"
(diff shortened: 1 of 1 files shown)


sample_apps/run_example.sh
```diff
@@ -407,20 +407,7 @@ function emded_policy_in_example_app() {
...
-        "simple_app_under_oe")
-            compile_simple_app_under_oe ""
-            ;;
-
-        *)
-            echo "${Me}: Unknown sample app name '${app_name}'."
-            return
-            ;;
-    esac
+    compile_"${SampleAppName}"
```
gapisback, line 410: The pattern of diffs in this change set is exactly of this nature:

- Rip out multiple lines of case / switch lines.
- Replace with single line calling to minion fn whose name is of the form : `<_verb_>_${app_name}`, or something to that effect.

This refactoring reduces the # of lines in this file and makes it easy to drop-in support for newer sample apps.

yelvmw (approved):
LGTM!

PR #80, opened 2023-05-20
Edit run_example.sh to consistently use 4-char tabstop.

gapisback:
This commit simply fixes up all tabs to consistently be 4-chars, and shiftwidth=4. No logic change in script.


gapisback:
@yelvmw - FYI; This is a simple edit-only change-set, where I've fixed all tabstops to be consistently 4 spaces in the shell script. Checking this in w/o reviews as no logic has changed.

PR #85, opened 2023-05-22
Update INSTALL.md

yelvmw:
Updated INSTALL.md to include optional Policy Generator dependencies and build instructions.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -93,6 +93,44 @@ $ make -f policy_utilities.mak clean
 $ make -f policy_utilities.mak
 ```
 
+If you wish to use the Policy Generator instead of manually composing the
+policies, you need to build it in utilities. The Policy Generator requires the
+json-schema-validator library which in turn depends on the JSON for Modern C++
+library. Follow the following instructions to have them built and installed on
+your system:
+
+```shell
+$ cd $(CERTIFIER)/utilities
+$ git clone https://github.com/nlohmann/json.git
+$ cd json
+$ mkdir build
+$ cd build
+$ cmake ..
+$ make
+$ sudo make install
+$ cd ..
+$ git clone https://github.com/pboettch/json-schema-validator.git
+$ cd json-schema-validator
+$ mkdir build
+$ cd build
+$ cmake .. -DBUILD_SHARED_LIBS=ON ..
+$ make
+$ sudo make install
+$ cd ..
+```
+
+Both libraries should be installed under /usr/local by default. If this is not
+the case, remember to update the JSON_VALIDATOR variable in
+policy_generator.mak. Add /usr/local/lib to /etc/ld.so.conf and run ldconfig if
+not already done.
+
+The Policy Generator utility can then be built using:
+
+```shell
+$ cd $(CERTIFIER)/utilities
+$ make -f policy_generator.mak
+```
+
 There is a Linux driver that simulates the SEV functions in sev-snp-simulator.
 Portions of this code are GPL licensed and the build driver is also GPL licensed.
 This is the only directory that contains GPL licensed code and it is not included in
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -93,6 +93,44 @@ $ make -f policy_utilities.mak clean
 $ make -f policy_utilities.mak
 ```
 
+If you wish to use the Policy Generator instead of manually composing the
+policies, you need to build it in utilities. The Policy Generator requires the
+json-schema-validator library which in turn depends on the JSON for Modern C++
+library. Follow the following instructions to have them built and installed on
```
gapisback, line 99: Maybe leave a blank line before the last sentence, and reword it as:

Follow these instructions ...

Otherwise, `Follow the following` sounds a bit odd to read.

INSTALL.md
```diff
@@ -93,6 +93,44 @@ $ make -f policy_utilities.mak clean
...
+$ cd build
+$ cmake .. -DBUILD_SHARED_LIBS=ON ..
+$ make
+$ sudo make install
+$ cd ..
+```
+
+Both libraries should be installed under /usr/local by default. If this is not
+the case, remember to update the JSON_VALIDATOR variable in
+policy_generator.mak. Add /usr/local/lib to /etc/ld.so.conf and run ldconfig if
```
gapisback, line 124: rephrase slightly as: "in the `policy_generator.mak` file."

gapisback, line 124: Try to consistently embed stuff in ``, like `/usr/local/lib`, `/etc/ld.so.conf` . Improves readability.

INSTALL.md
```diff
@@ -93,6 +93,44 @@ $ make -f policy_utilities.mak clean
 $ make -f policy_utilities.mak
 ```
 
+If you wish to use the Policy Generator instead of manually composing the
```
gapisback, line 96: Here, do we need to clarify that " ... use the Policy Generator for building the `simple_app_under_oe` ...".

Am not sure if this generator will be used for / by other apps. So, it's clearer to narrow its usage to the one app that needs it.

INSTALL.md
```diff
@@ -93,6 +93,44 @@ $ make -f policy_utilities.mak clean
...
+```shell
+$ cd $(CERTIFIER)/utilities
+$ git clone https://github.com/nlohmann/json.git
+$ cd json
+$ mkdir build
+$ cd build
+$ cmake ..
+$ make
+$ sudo make install
+$ cd ..
```
gapisback, line 111: Maybe leave a blank line before and after this line. And similarly for the `cd ..` on L119 below.

That way it will be clear that the prev set of commands are related to one repo. Blank line will visually separate out the work done for one repo from that done for the 2nd repo.

gapisback (approved):
@yelvmw - Now that you've pulled-in the instructions to the main INSTALL.md file, do we need to delete this file?

https://github.com/vmware-research/certifier-framework-for-confidential-computing/blob/generator_doc/utilities/policy_generator_info.txt

It still has some more content that may be useful and may need to be retained. If so, it's ok to leave it around. You may just want to log an issue if any follow-on clean-up for this `policy_generator_info.txt` is still needed; i.e. either rm it or re-write it in md-format w/ reduced content.

For now, what you have is good enough to go ahead.

INSTALL.md
yelvmw: Policy Generator works with any backend. Not just OE. For instance, SEV-SNP sample uses it too.

INSTALL.md
yelvmw: Done

INSTALL.md
yelvmw: Done

yelvmw: Done

INSTALL.md
yelvmw: Done

PR #103, opened 2023-06-12
Keystone certifier

jlmucb:
Initial Keystone port


src/keystone/shim_test.mak
```diff
@@ -0,0 +1,106 @@
...
+dobj = $(O)/certifier.pb.o $(O)/certifier.o $(O)/certifier_proofs.o        \
+       $(O)/support.o $(O)/application_enclave.o $(O)/simulated_enclave.o  \
+       $(O)/cc_helpers.o $(O)/cc_useful.o $(O)/keystone_shim.o $(O)/keystone_test.o
+
+all:	$(CL)/keystone_test.exe
+clean:
+	@echo "removing object files"
+	rm -rf $(O)/*.o
+	@echo "removing executable files"
+	rm -rf $(CL)/keystone_test.exe
```
gapisback, line 61: This `*.exe` is being built in current dir, `keystone/keystone_test.exe`.

Shouldn't `clean` rm from current dir rather than from `$(CL)` ???

gapisback: And, as exe generates the 2 `*.bin` files, can we have `clean` rm those too?

That way, we are sure to be starting from a clean slate w/o older .bin files hanging around.

src/keystone/keystone_shim.cc
```diff
@@ -0,0 +1,346 @@
+#include "keystone_api.h"
+#include "certifier_framework.h"
+#include "certifier_utilities.h"
+#include <string.h>
+
+using std::string;
+using namespace certifier::utilities;
+
+
+//  Copyright (c) 2021-22, VMware Inc, and the Regents of the University of California.
```
gapisback, line 10: This shim file may go away eventually ... but, meanwhile, for consistency, can you move this to the top of the file?

gapisback (approved):
Took a quick look at all the files, to get familiar with this shim approach.

Have not verified that the instructions work, but will figure those out when integrating this under `run_example.sh`, for issue #106.

For now, this is good to go, to enable downstream development.

yelvmw (approved):
This change set is large. I tried to cover some of the key areas only. So far, it looks good. I think we can upstream after Aditya's comments are addressed.

PR #104, opened 2023-06-12
Filter policy

jlmucb:


jlmucb:
This supports mutiple measurements and platforms in certifier policy including certifier services with heterogeneous Confidential Computing platforms (e.g.-AMD SEV and Arm CCA).

PR #110, opened 2023-06-16
Initial apply of 'gofmt' with default formatting rules.

gapisback:
Draft PR to do self-review, and see if CI still passes.

--- Email contents ----

I looked around and found [the tool, gofmt](https://go.dev/blog/gofmt), which seems the standard formatting tool for Go.

$ go doc cmd/gofmt says:

---
Gofmt formats Go programs. It uses tabs for indentation and blanks for
alignment. Alignment assumes that an editor is using a fixed-width font.

Unfortunately, they don’t seem to encourage or support site-specific customization using, e.g, gofmt format-files.
There is some mention of [such interfaces](https://www.tutorialspoint.com/auto-format-go-programming-language-source-code-with-gofmt) on the net, but I could not get them to work.

### DIFF
--- a/certifier_service/certlib/cert1_test.go
+++ b/certifier_service/certlib/cert1_test.go
@@ -15,1252 +15,1247 @@
 package certlib
 
 import (
-        "bytes"
-        "crypto"
-        "crypto/elliptic"
-        "crypto/ecdsa"
-        "crypto/sha256"
-        "crypto/sha512"
-        "crypto/rand"
-        "crypto/rsa"
-        "crypto/x509"
-        "crypto/x509/pkix"
-        "fmt"
-        "io"
-        "math/big"
-        //"net"
-        "os"
-        //"syscall"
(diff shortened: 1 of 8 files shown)


gapisback:
HI, @jlmucb & @yelvmw -- Here is a draft PR just for inspection and evaluation.

The `gofmt` tool automatically implements some default formatting:

- using tabs for indentation.
- removes unnecessary `( )` and other formatting characters
- re-formats some expressions, in some cases, to remove white-space which appear superfluous

Several other styles, which are 'default' as per Go-formatting rules.

This change will require some adjustments in our personal workflows (use of `set tabstop=4`, settings for `shiftwidth` etc.)

Overall, it does make the code formatting consistent and readable.

We can consider this change after a couple of weeks, once our dev-dust settles.

gapisback:
This work has been re-done and finalized under PR #125

Issue #111, opened 2023-06-16
Improve error reporting from CC utility programs, to flag invalid inputs, missing files etc.


gapisback:
During development of `run_example.sh`, script errors are silently ignored by some utility programs. This is misleading and causes user-errors to go undetected, w/o printing error messages.

Here are some examples:

Following fails with non-zero $rc, but no message is printed: `--input` file's path is wrong here, so policy file is not found.

```
cc-sdb-vm:[19] $ /home/agurajada/Code/certifier-framework-for-confidential-computing/utilities/print_packaged_claims.exe --input=policy.bin
cc-sdb-vm:[20] $ echo $?
1
```

The following incorrect usage, specifying a wrong measurement file, `sample-example_app.measurement`, is correctly reporting an error message and non-zero $rc:

```
$ /home/agurajada/Code/certifier-framework-for-confidential-computing/sample_apps/simple_app/example_app.exe --data_dir=./app2_data/ --operation=cold-init --measurement_file=sample-example_app.measurement --policy_store_file=policy_store --print_all=true
simulated_Init: Can't get measurement file size ./app2_data/sample-example_app.measurement
initialize_simulated_enclave_data: simulated_init failed
Can't init simulated enclave

cc-sdb-vm:[32] $ echo $?
1
```

However, an error in the policy-file (see: `--policy_store_file=policy_store-Wrong-One`) is not being detected. $rc is still 0.

This creates difficulties while scripting these commands, while integrating support for new sample-apps.

```
$ /home/agurajada/Code/certifier-framework-for-confidential-computing/sample_apps/simple_app/example_app.exe --data_dir=./app2_data/ --operation=cold-init --measurement_file=example_app.measurement --policy_store_file=policy_store-Wrong-One --print_all=true

cc-sdb-vm:[36] $ !echo
echo $?
0
```

PR #113, opened 2023-06-18
(#111) Fix checks to report error messages when input args are invalid.

gapisback:
This commit tightens some error checking of invalid arguments to few utility programs that were silently failing without reporting error messages.

No new error checks are added; only error messages are being printed when needed.

### DIFF
--- a/src/cc_helpers.cc
+++ b/src/cc_helpers.cc
@@ -242,7 +242,14 @@ bool certifier::framework::cc_trust_data::initialize_keystone_enclave_data(const
       printf("initialize_keystone_enclave_data: Not a simulated enclave\n");
       return false;
     }
-    // Todo
+    int m_size = file_size(measurement_file_name);
+    if (m_size < 0) {
+        printf("%s(): Invalid size=%d of measurement file '%s'.\n",
+               __func__, m_size, measurement_file_name.c_str());
+        return false;
+    }
+
+    // TODO
     byte der_cert[100];
     if (!keystone_Init(0, der_cert)) {
       printf("initialize_keystone_enclave_data: keystone_init failed\n");
--- a/utilities/embed_policy_key.cc
+++ b/utilities/embed_policy_key.cc
@@ -85,7 +85,11 @@ bool read_file(string file_name, int* size, byte* data) {
 bool generate_policy_cert_in_code(string& asn1_cert_file, string& include_file) {
   int cert_size = file_size(asn1_cert_file);
   if (cert_size <= 0)
+  {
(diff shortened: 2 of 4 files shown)


src/cc_helpers.cc
```diff
@@ -242,7 +242,14 @@ bool certifier::framework::cc_trust_data::initialize_keystone_enclave_data(const
       printf("initialize_keystone_enclave_data: Not a simulated enclave\n");
       return false;
     }
-    // Todo
+    int m_size = file_size(measurement_file_name);
+    if (m_size < 0) {
```
gapisback, line 246: This was just not being checked. This missing check caused me tons of time wasted when porting the Keystone-shim based sample app to run under `run_example.sh` .

The generated measurement file was named slightly differently than the default name used by other apps; i.e. `example_app.measurement`. The initial instructions had named it `keystone_example_app.measurement`, which caused the file to not-be-found and some parts of the workflow would silently exit.

This new error message will improve the situation in future when others attempt to use the workflow and make a similar mistake.

utilities/embed_policy_key.cc
```diff
@@ -85,7 +85,11 @@ bool read_file(string file_name, int* size, byte* data) {
 bool generate_policy_cert_in_code(string& asn1_cert_file, string& include_file) {
   int cert_size = file_size(asn1_cert_file);
   if (cert_size <= 0)
+  {
+    printf("Invalid size=%d for input file '%s'.\n",
```
gapisback, line 89: Add an error message rather than returning silently.

utilities/print_signed_claim.cc
```diff
@@ -30,7 +30,7 @@ bool get_signed_from_file(const string& in, signed_claim_message* sc) {
   byte serialized_cm[in_size];
 
   if (!read_file(in, &in_read, serialized_cm)) {
-    printf("Can't read %s\n", in.c_str());
+    printf("Can't read input file '%s'.\n", in.c_str());
```
gapisback, line 33: Make error message a bit more accurate & descriptive.

utilities/print_signed_claim.cc
```diff
@@ -48,7 +48,7 @@ int main(int an, char** av) {
 
   signed_claim_message sc;
   if (!get_signed_from_file(FLAGS_input, &sc)) {
-    printf("Can't get signed claim fro, file\n");
+    printf("Can't get signed claim from file '%s'.\n", FLAGS_input.c_str());
```
gapisback, line 51: Fixes typo ...

yelvmw (approved):
LGTM!

Issue #116, opened 2023-06-21
Reconcile 2 copies of ./src/certifier.proto and ./certifier_service/certprotos/certifier.proto . Duplicate file causes build confusion.


gapisback:
In `/main`, we see these two `certifier.proto` files which are nearly identical. 
```
agurajadaHMD6R:[41] $ !diff
diff ./src/certifier.proto ./certifier_service/certprotos/certifier.proto
18c18
< //option go_package = "github.com/jlmucb/crypto/v2/certifier_prototype/certifier_service/certprotos";
---
> option go_package = "github.com/jlmucb/crypto/v2/certifier_prototype/certifier_service/certprotos";
```

Presence of these duplicate files creates build-confusion, and tends to be error prone.

Figure out the source of this duplication, and resolve by purging one of these files. Most likely, `./certifier_service/certprotos/certifier.proto` may be the more intuitive file to retain.

gapisback:
From @jlmucb : it is possible that if you uncomment the go_package stuff and it compiles in the C-code, then you don't have ot have to versions. I.e. Run `protoc` ... as it's being in the build instructions. Try this out .. and if things work, John is ok by it.

PR #121, opened 2023-06-23
(#116) Purge src/certifier.proto in favour of certprotos version. Adjust Makefiles in all sub-dirs.

gapisback:
This commit cleans up duplicate versions of proto-files, which live(d) in
- ./src/certifier.proto  (now deleted)
- ./certifier_service/certprotos/certifier.proto (retained)

Duplicate files created unnecessary confusion / duplication of dev- changes. 

All affected *.mak and Makefile's are updated to consistently use version from `$(CP)/certifier.proto`, where `CP` is defined to be `$(CERTIFIER_ROOT)/certifier_service/certprotos` .

The build steps for Simple-apps under Gramine and OE have been manually
tested on SGX-machine. simple_app_under_sev/ is tested under CI, so
that should also build correctly.

- Minor corrections to run_example.sh to improve help messages
- Updates to build.yml to optimize run for simple_app_under_app_service/
---

**NOTE**: The choice of retaining the file in certprotos/ subdir is because it's also used in `run_example.sh` to invoke `go build simpleserver.go`. This `certifier_service/certprotos/certifier.proto` file is referenced in all instructions.md files. So, if we had to chosen to delete this file in preference to the one in `src/` dir, it would involve updates to many more instructions.md file(s). With the choice currently made, all changes are hidden behind `.mak`-files.

**NOTE**: The consistent pattern I have tried to apply to all Makefiles affected is as follows:

- Top-level symbol for `CERTIFIER_ROOT = ../..` (as appropriate)
- Add: `CP = $(CERTIFIER_ROOT)/certifier_service/certprotos`
- Use `$(CP)/certifier.proto` to locate the proto-file
- Use `--proto_path=$(CP)` flag for `proto` (in favour of `-I` flag) as I felt `proto_path` name is more intuitive.
- Most places use `$(US)` to name the current source-dir (i.e. usually "./") Some Make-files may have `$(S)` for that
- Add changes in `clean` target to : `rm -rf $(US)/certifier.pb.cc $(US)/certifier.pb.h $(I)/certifier.pb.h`

I have tried to consistently apply this pattern to all Makefiles affected, so that it will be easier for future readers to understand where the proto-file is being used from, and where it's generated etc.


.github/workflows/build.yml
```diff
@@ -171,7 +171,7 @@ jobs:
         echo " "
         echo "---- simple_app_under_app_service/ setup ----"
         echo " "
-        ./run_example.sh simple_app_under_app_service setup
+        ./run_example.sh --no-make-clean simple_app_under_app_service setup
```
gapisback, line 174: We've already built the utilities etc. in previous step of `./run_example.sh application_service setup`.

So we can now skip re-building those using this new option, added in a previous change.

jlmucb (approved):

PR #132, opened 2023-06-27
Fix Policy Generator bug

yelvmw:
Fixed policy generator bug when multiple measurements and multiple claims are present.


gapisback (approved):
fix seems reasonable from a quick-check.

Are there any tests we can enable in CI to verify this fix, if we integrate this to /main now?

I'm assuming you ran into this bug while fixing the filter-policy scheme, ,right? So, this fix is not needed right now in our branch until the time PR #104 comes-in, yes? If so, does it make sense to bundle this fix along with the change set for that other PR ?

Issue #141, opened 2023-07-05
Develop support for Python bindings to use Certifier APIs.


gapisback:
Implement support for Python bindings to develop apps using Certifier APIs.

Use the SWIG s/w tool to generate Python bindings using Certifier headers.

PR #144, opened 2023-07-07
(#141) Implement SWIG-generated Python bindings for Certifier Framework

gapisback:
This commit adds SWIG-interfaces to generate Python bindings to the Certifier Framework and Certifier tests. 

This commit adds sufficient scaffolding to add these Python bindings. (Future work will flesh this
out to build a standalone simple-app written in Python.)

Make-file changes are done to invoke swig, and to generate 2 shared libraries:

- `libcertifier_framework.so` - Core Certifier library APIs
- `libcertifier_tests.so` - Wrappers for most tests in certifier_tests.cc

Add pytest machinery to exercise these Python bindings, as a way to validate the build process, and to verify that these Python bindings basically do work to invoke the Certifier code.

Add these two pytest-based test driver files:

- test_cert_framework_basic.py - Tests for basic validation of Python bindings to core Certifier Framework APIs.
- test_certifier_tests.py - Tests to validate coherence of `libcertifier_tests.so`. And, then, to exercise all existing tests
  defined in `certifier_tests.cc` (Google tests) through their Python wrapper methods.

Incorporate these changes to CI build.yml, with appropriate flags, so that we can see the output from certifier_tests when executed thru pytest.

----

**NOTE to the reviewers**: Bulk of code is ready to roll. During review, we should evaluate and discuss:

1. Use of `CFLAGS=-fPIC` for building files for the shared lib v/s for normal `.exe` builds. Make-files need some revamping. The issue here is that files that go into the shared library need to be compiled with `-fPIC` flag, but that's really not needed for normal executables, or the `.a`. Question is : Would it be ok to just compile all files with `-fPIC`, so we do only one build. And use the objects either for the `.exe` or for the shared library. [ Currently, things seem to work ... but I need to investigate the build-time requirements here and resolve one way or the other. ]

2. Location of generated `*.so` libraries. Currently they come at top-level certifier-root dir
3. Location of new `tests/` dir. I think a sibling of `src/` is the right place; but am open to suggestions
4. Do we need `tests/pytests/` to clarify what kind of tests these are? Then, in future, all existing `_tests.cc` could go into `tests/` subdir. (I prefer this dir-layout, as it cleans-up dir-location for `src/` v/s pure `tests/*_tests.cc` files.)

**NOTE on CI-run-times**:   This is the Cost of doing proper s/w engineering business.

With the addition of Swig-based builds, and the new pytest tests, E2E CI times have gone up from about 16-17 mins to ~18-22 mins. (Varies based on CI-build machines' load.)

Some of the gains we made recently with introduction of parallel make have been eroded by the few extra builds needed to juggle the Swig-builds. And then to re-run Certifier tests thru pytest framework.


.github/workflows/build.yml
```diff
@@ -13,21 +13,24 @@ jobs:
...
+    #! -------------------------------------------------------------------------
     - name: dependencies
       run: |
         sudo apt update -y
         sudo apt install -y libgtest-dev libgflags-dev openssl libssl-dev protobuf-compiler protoc-gen-go golang-go cmake
+        sudo apt install -y python3 pylint
+        pip install pytest
 
+        # Needed for Python bindings support
+        sudo apt install -y swig
```
gapisback, line 25: Introduce `swig` in CI ... and a collection of changes to invoke `make`, run `pylint` ... etc ...

Basically, bunch of CI integration.

.github/workflows/build.yml
```diff
@@ -39,13 +42,64 @@ jobs:
...
+        # We need to clean here, otherwise make certifier_tests.mak will run
+        # into some protobuf-related errors. But this 'clean' will also
+        # rm ../libcertifier_framework.so, so it will need to be re-built.)
         make -f certifier.mak clean
 
         make -j2 -f certifier_tests.mak
         ./certifier_tests.exe
 
+        # Re-build ...; otherwise pylint will complain below.
+        make -f certifier.mak ../libcertifier_framework.so
```
gapisback, line 55: Minor wart in our make systems ... difficult to resolve otherwise, w/o extra effort.

This should be ok for now..

.github/workflows/build.yml
```diff
@@ -39,13 +42,64 @@ jobs:
...
+    #! where the shared library (imported by Pytests code) is built.
+    #! -------------------------------------------------------------------------
+    - name: test-code-formatting
+      run: |
+        ./CI/scripts/check-gofmt.sh
+
+        pushd tests
+
+        set -x
+        pylint *.py
```
gapisback, line 74: Hard check now onwards. All new pytest code must clear pylint.

.github/workflows/build.yml
```diff
@@ -39,13 +42,64 @@ jobs:
...
+        pushd tests
+
+        # Run one case showing verbose outputs of each library's contents.
+        # Run each py file separately as 'pytest -v' (on all files) craps out.
+        set -x
+        pytest --capture=tee-sys test_cert_framework_basic.py -k test_getmembers_of_cc_framework
+
+        PYTHONUNBUFFERED=TRUE pytest -v test_cert_framework_basic.py
+
+        PYTHONUNBUFFERED=TRUE pytest --capture=tee-sys -v test_certifier_tests.py
```
gapisback, line 97: We need to do this `PYTHONUNBUFFERED=TRUE` stuff so that the output from pytest, reporting each test case being run, is clearly interleaved with output generated by the test itself.

.gitignore
```diff
@@ -29,3 +29,10 @@ enclave/enclave
 # Ignore ISLET-related download and build artifacts
 *shim_test
 *.tar.gz
+
+# Ignore files generated by Python bindings' build process(es)
+__pycache__
+certifier_framework_wrap.cc
+certifier_tests_wrap.cc
+libcertifier_framework.py
+libcertifier_tests.py
```
gapisback, line 38: All generated files. I think these need not be committed.

We can change our minds once we get better experience with the build processes and toolchain, once simple-app in Py gets developed.

include/certifier_tests.h
```diff
@@ -0,0 +1,28 @@
...
+
+// This list should be repeated in the same order in certifier_tests.i,
+// which defines the SWIG interface for the Python bindings for the
+// certifier_tests.so shared library.
+
+#include "primitive_tests.h"
+#include "claims_tests.h"
+#include "store_tests.h"
+#include "x509_tests.h"
+#include "support_tests.h"
```
gapisback, line 26: Had to carve out extern prototypes for certifier test methods into these `.h` files, so that these get picked up by the swig interfaces files.

include/claims_tests.h
```diff
@@ -0,0 +1,31 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef __CLAIMS_TESTS_H__
+#define __CLAIMS_TESTS_H__
+
+// RESOLVE: Need this? using namespace certifier::framework;
```
gapisback, line 18: @yelvmw -- I think these namespaces are not needed by these tests. Things worked just fine w/o these namespaces.

Let me know ... RESOLVE can probably go away.

src/certifier.cc
```diff
@@ -371,6 +371,10 @@ bool certifier::framework::policy_store::Deserialize(string& in) {
   return true;
 }
 
+int certifier::framework::policy_store::get_num_signed_claims() {
```
gapisback, line 374: This interface was missing previously. Swig-bundling complains ... so it was good to find and fix this missing piece, as a side-benefit of trying out this Swig business.

src/certifier.mak
```diff
@@ -41,18 +41,26 @@ INCLUDE=-I $(I) -I/usr/local/opt/openssl@1.1/include/ -I $(S)/sev-snp
...
 LINK=g++
 
 # Point this to the right place, if you have to, based on your machine's install:
 # PROTO=/usr/local/bin/protoc
 PROTO=protoc
 AR=ar
+
+# Definitions needed for generating Python bindings
+SWIG=swig
+PY_INCLUDE = -I /usr/include/python3.10/
```
gapisback, line 62: This hard-coded path ... is generally not good. But I found that this seems to be the current-default rev-level.

It's working in CI, too.

Any suggestions on how-to create this -I definition using some symbols, to future-proof this area?

src/certifier.mak
```diff
@@ -68,18 +76,34 @@ ifdef ENABLE_SEV
...
+# Objs needed to build Certifer Framework shared lib for use by Python module
+cfsl_dobj := $(dobj) $(O)/certifier_framework_wrap.o
+
+CERTIFIER_LIB = certifier.a
+CERTIFIER_SHARED_LIB = libcertifier_framework.so
+
+# RESOLVE: We are invoking shared lib target first, which means
+# CFLAGS_PIC = -fpic will be included in CFLAGS for all objects.
+# Rearrange so that for normal builds we don't include CFLAGS_PIC = -fpic
+# and only re-build all required objects with -fpic for shared library's build.
```
gapisback, line 88: RESOLVE: Need to think how to fix this issue.

src/certifier.mak
```diff
@@ -98,6 +122,20 @@ $(O)/certifier.o: $(S)/certifier.cc $(I)/certifier.pb.h $(I)/certifier.h
 	@echo "\ncompiling $<"
 	$(CC) $(CFLAGS) -o $(@D)/$@ -c $<
 
+$(O)/certifier_pic.o: $(S)/certifier.cc $(I)/certifier.pb.h $(I)/certifier.h
```
gapisback, line 125: Note to myself: This rule can go away; I think.

src/certifier_tests.cc
```diff
@@ -38,158 +38,153 @@ DEFINE_string(ask_cert_file_name, "ask_cert_file.bin", "ask cert");
 DEFINE_string(vcek_cert_file_name, "vcek_cert_file.bin", "vcek cert");
 
 // Encryption and support tests
-extern bool test_random(bool print_all);
```
gapisback, line 41: All these deleted `extern bool` prototypes are now replaced, properly, with `#include .h` files, where-in these prototypes are defined.

src/certifier_tests.cc
```diff
@@ -38,158 +38,153 @@ DEFINE_string(ask_cert_file_name, "ask_cert_file.bin", "ask cert");
...
   EXPECT_TRUE(test_claims_1(FLAGS_print_all));
 }
-extern bool test_signed_claims(bool print_all);
+
 TEST (signed_claims, test_signed_claims) {
   EXPECT_TRUE(test_signed_claims(FLAGS_print_all));
 }
 
-extern bool test_local_certify(string&, bool, string&, string&);
+extern bool test__local_certify(string&, bool, string&, string&);
```
gapisback, line 111: This `test_local_certify()` and `test__local_certify()` jugglery is just so that we have the former available with the default signature (i.e., `(bool print_all)`), just like all the other test methods.

The former variant becomes the default test-case method, with default arguments, supplied to call the underlying latter method.

Then the former, default, test-case method gets called via the Python bindings from `test_certifier_tests.py()`

src/certifier_tests.mak
```diff
@@ -89,20 +104,27 @@ channel_dobj += $(sev_common_objs)
 pipe_read_dobj += $(sev_common_objs)
 endif
 
-all:	certifier_tests.exe test_channel.exe pipe_read_test.exe
+# RESOLVE: Same issue with ordering and use of CFLAGS_PIC = -fpic for all objects.
```
gapisback, line 107: Note to myself: Fix this RESOLVE issue ...

.github/workflows/build.yml
```diff
@@ -39,13 +42,64 @@ jobs:
...
+        pushd tests
+
+        set -x
+        pylint *.py
+        set +x
+
+        popd
+
+    #! -------------------------------------------------------------------------
+    - name: test-cert_framework-pytests
```
gapisback, line 80: See the output for this chunk `test-cert_framework-pytests` in the CI runs, to get an idea exactly-how these pytest tests run, and what outputs they will create.

tests/test_certifier_tests.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    Execute all Certifier tests from the shared library.
+    This test-case drives execution of all C++ unit-tests in one single swoop.
+    """
+    cft_builtin_methods = getmembers(libct, isbuiltin)
+
+    for item in cft_builtin_methods:
+        print('\n**** Execute pytest: ', item[0], ' ****\n', flush=True)
+        # pylint: disable-next=exec-used
+        exec('libct.' + item[0] + '(True)')
```
gapisback, line 68: I have written this test method in a (somewhat obscure) way so that it's future-proof'ed to automatically include, inspect and execute any test case that may be added to `certifier_tests.cc`. 

This way, we get **automatic coverage** both from C++ `certifier_tests.exe` and then **those same tests** are executed from this Python interface.

.github/workflows/build.yml
gapisback: Note to myself: Update comment to read: 

"Rebuild shared library as previous make clean deleted it. Otherwise, pylint, below, will complain that .so was not found."

jlmucb (approved):
Looking at the code I now understand why you want a test directory.

include/certifier_framework.h
```diff
@@ -145,9 +145,11 @@ namespace certifier {
 
     class cc_trust_data {
 
-    static const int max_symmetric_key_size_ = 128;
```
yelvmw, line 148: I remember the reason for this change. Are there other solutions?

include/claims_tests.h
yelvmw: Did the tests already use the namespaces? It's probably OK to just ignore it.

src/certifier.mak
yelvmw: Not an expert here. But this can't be good. My SGX machine for instance does not have 3.10. Let's do some digging.

yelvmw (approved):
Largely looking good. I have some comments. My main concern is the Python version issue. The -fPIC problem is probably OK to defer to a later time.

.github/workflows/build.yml
gapisback: Updated comment.

include/claims_tests.h
gapisback: Took these comments out... Not specifying namespaces seemed to not make a diff for the builds and tests.

.github/workflows/build.yml
gapisback: Update: In final version, relocated new Python tests to `tests/pytests` sub-dir and adjusted commands in build.yml that run `pylint` or execute `pytest`.

include/certifier_framework.h
gapisback: @yelvmw - I don't think there are other solutions. I tried to redefine this limit using an enum-constant, but even that definition ran into issues.

Here is an example of a compilation failure I got with one of the alt solutions attempt:

```
../include/certifier_framework.h:149:50: error: ‘MAX_SYMMETRIC_KEY_SIZE’ was not declared in this scope
  149 |       static const int max_symmetric_key_size_ = MAX_SYMMETRIC_KEY_SIZE;
      |                                                  ^~~~~~~~~~~~~~~~~~~~~~
```

Here's another example,

Both are defined as private:

```
148       enum {MAX_SYMMETRIC_KEY_SIZE = 128};
149       // Python swig bindings need this to be public, to size other array decls
150       static const int max_symmetric_key_size_ = MAX_SYMMETRIC_KEY_SIZE;
```

Compilation fails like:

```
certifier_framework_wrap.cc: In function ‘PyObject* _wrap_cc_trust_data_symmetric_key_bytes__set(PyObject*, PyObject*)’:
certifier_framework_wrap.cc:7149:64: error: ‘certifier::framework::cc_trust_data::<unnamed enum> certifier::framework::cc_trust_data::MAX_SYMMETRIC_KEY_SIZE’ is private within this context
 7149 |       for (; ii < (size_t)certifier::framework::cc_trust_data::MAX_SYMMETRIC_KEY_SIZE; ++ii) *(byte *)&arg1->symmetric_key_bytes_[ii] = *((byte *)arg2 + ii);
      |                                                                ^~~~~~~~~~~~~~~~~~~~~~
In file included from certifier_framework_wrap.cc:2824:
../include/certifier_framework.h:148:13: note: declared private here
  148 |       enum {MAX_SYMMETRIC_KEY_SIZE = 128};
      |             ^~~~~~~~~~~~~~~~~~~~~~
certifier_framework_wrap.cc: In function ‘PyObject* _wrap_cc_trust_data_service_symmetric_key__set(PyObject*, PyObject*)’:
certifier_framework_wrap.cc:7715:64: error: ‘certifier::framework::cc_trust_data::<unnamed enum> certifier::framework::cc_trust_data::MAX_SYMMETRIC_KEY_SIZE’ is private within this context
 7715 |       for (; ii < (size_t)certifier::framework::cc_trust_data::MAX_SYMMETRIC_KEY_SIZE; ++ii) *(byte *)&arg1->service_symmetric_key_[ii] = *((byte *)arg2 + ii);
      |                                                                ^~~~~~~~~~~~~~~~~~~~~~
```

include/certifier_tests.i
```diff
@@ -0,0 +1,26 @@
+// certifier_tests.i- SWIG interface to generate Python bindings to Certifier unit tests
+// *****************************************************************************
+// A large-ish collection of Certifier tests, defined in certifier_tests.cc and
+// related *_tests.cc files, are exposed through Swig generated Python bindings.
+// All those Certifier tests are invoked thru pytest, test_certifier_tests.py .
+// This interface file makes that glue possible through the build process(es).
+// *****************************************************************************
+
+%module libcertifier_tests
+%include "std_string.i"
```
gapisback, line 10: Deleted this line. Discovered that it's not needed in this `.i` file but IS needed in `include/certifier_framework.i`

src/certifier_tests.mak
gapisback: Resolved. Added new target `sharedlib` and make command will need to be invoked using `-B` or `--always-make` argument when building this target.

Updated commands in `build.yml` to do a 2nd build to build the shared libraries.

src/certifier.mak
gapisback: Resolved. Added `sharedlib` as new target and changed build.yml to invoke `make` with `--always-make` flag.

PR #150, opened 2023-07-12
Relocate Copyright notice to top-of-file in many .h/.cc files.

gapisback:
This commit fixes many `.h` and `.cc` files to relocate the Copyright notice to the top-of-the file. 
No further code / logic changes are included.


gapisback:
HI, @yelvmw - While working on Python bindings stuff ... I keep bumping into files where the copyright notice is not at the right place.

I made a sweep earlier, and here's a 2nd sweep to clean many files. This review is fyi ... If you don't mind, I'd like to check this into /main as soon as the tests pass. (Then, I can rebase on top of this change-set in my Py-bindings dev-work, and not be continually distracted by these mis-placed Copyright notices.)

PR #151, opened 2023-07-15
New policy store

jlmucb:
New policy store is more compact and provides more general capabilities.
New policy store will be needed for multi-domain certification.
Took the opportunity to fix some names that didn't conform to standard syntax.

Works with simple_app


src/certifier.cc
```diff
@@ -25,444 +25,212 @@ using namespace certifier::utilities;
...
-    return false;
-  tagged_claim* t = new(tagged_claim);
-  int n = num_tc_++;
-  t->set_tag(tag);
-  tc_[n] = t;
-  (tc_[n]->mutable_claim())->CopyFrom(to_add);
-  return true;
+void certifier::framework::policy_store::get(int ent, string* v) {
+  if (ent < 0 || ent >= num_ents_)
+    return;
```
gapisback, line 113: Change interface to `return false` here.

src/certifier.cc
```diff
@@ -25,444 +25,212 @@ using namespace certifier::utilities;
...
+  *v = entry_[ent]->value_;
 }
 
-void certifier::framework::policy_store::delete_claim_by_index(int n) {
-  if (n >= num_tc_)
+void certifier::framework::policy_store::put(int ent, const string v) {
+  if (ent < 0 || ent >= num_ents_)
```
gapisback, line 118: Check for `ent < 0` will not be needed if you changed to `unsigned ent`.

Change interface to `return false` here, if this check fails.

include/certifier_framework.h
```diff
@@ -34,87 +34,60 @@ using std::string;
...
-
+      policy_store(int max_ents);
       policy_store();
-      policy_store(const string enc_alg, int max_trusted_services, int max_trusted_signed_claims,
-          int max_storage_infos, int max_claims, int max_keys, int max_blobs);
       ~policy_store();
 
-      bool replace_policy_key(key_message& k);
-      const key_message* get_policy_key();
+      int get_num_entries();
```
gapisback, line 70: Consider returning `unsigned` ... for consistency with other interfaces that receive `unsigned ent` as the index-of-an-entry argument.

include/certifier_framework.h
```diff
@@ -34,87 +34,60 @@ using std::string;
...
-      void delete_authentication_key_by_index(int index);
-
-      bool add_blob(const string& tag, const string& s);
-      const string* get_blob_by_tag(const string& tag);
-      const string* get_blob_by_index(int index);
-      const tagged_blob_message* get_tagged_blob_info_by_index(int n);
-      int get_blob_index_by_tag(const string& tag);
-      void delete_blob_by_index(int index);
-      int get_num_blobs();
+      bool update_or_insert(const string& tag,
```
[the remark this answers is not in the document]
gapisback: Looking at your current implementation for `update_or_insert()`, I think if you just have `upsert()` as the single  **public** interface and take away `add_entry()`, the uniqueness check problems will likely be solved on their own.

- `add_entry()` still continues to live as a private interface.
- `upsert` will call `find_entry_by_tag_and_type()`, and if found, it will update the value.
- Otherwise, add a new entry

Currently, these APIs do not seem to support concurrency, so above logic will suffice to ensure uniqueness checks on (tag, type).

include/certifier_framework.h
```diff
@@ -34,87 +34,60 @@ using std::string;
...
+              const string& type, const string& value);
 
+      const key_message* get_policy_key();
+      bool set_policy_key(const key_message key);
+
+      void print();
```
gapisback, line 87: This one is fine, but the following two `Serialize` and `Deserialize` are named with upper-case letters which is inconsistent with the naming of the rest of the methods.

Why not name these as `serialize()` and `deserialize()`? I think that will be consistent w/ other public-facing methods.

src/cc_helpers.cc
```diff
@@ -505,60 +505,80 @@ void certifier::framework::cc_trust_data::clear_sensitive_data() {
   // Not necessary on most platforms.
 }
 
+//  cc_trust_data relies on the following data in the store
+//    public_policy_key
+//    public_key_algorithm
+//    symmetric_key_algorithm
+//    for attestation:
```
gapisback, line 512: Minor: Change to 'For'

src/cc_helpers.cc
```diff
@@ -505,60 +505,80 @@ void certifier::framework::cc_trust_data::clear_sensitive_data() {
...
+//      service-attest-key
+//      sealing-key
+//    for authentication
```
gapisback, line 515: Minor: change to:  For authentication:

('F' and trailing ':')

src/cc_helpers.cc
```diff
@@ -505,60 +505,80 @@ void certifier::framework::cc_trust_data::clear_sensitive_data() {
...
+//      auth-key
+//      app-symmetric-key
+//    Some platforms require obtaining a platform rule. For them, it should
```
gapisback, line 518: In the phrase, "In them, it ..." ... the word 'it' is ambiguous. How about rephrasing this as:

"platform-rule (signed-claim) is also stored for platforms that require a platform-rule."

Is that clearer?

src/cc_helpers.cc
```diff
@@ -505,60 +505,80 @@ void certifier::framework::cc_trust_data::clear_sensitive_data() {
...
+//    be stored.  platform-rule (signed-claim)
+//  initialized cert is in the array initialized_cert with size initialized_cert_size
```
gapisback, line 520: Not sure about this new comment line.

I did not find any references to 'initialized_cert', or 'initialized_cert_size` in `certifier_framework.h`.

What are we trying to refer to here with this comment? Might need some corrections / clarifications.

gapisback (changes requested):
HI, @jlmucb -- 

I finished my 1st pass on the reviewing going through core interface changes in `certifier_framework.h` and some of the associated implementations in `certifier.cc`.

I'm pushing this out now so you can see this first, while I go thru the diffs in the other files / `.cc` changes.

I think we need to talk through your proposed interface changes in a call, to finalize some of my comments / concerns listed here.

Thanks.

include/certifier_framework.h
```diff
@@ -34,87 +34,60 @@ using std::string;
...
-      bool replace_policy_key(key_message& k);
-      const key_message* get_policy_key();
+      int get_num_entries();
+      bool add_entry(const string& tag, const string& type, const string& value);
+      int find_entry_by_tag(const string& tag);
+      int find_entry_by_tag_and_type(const string& tag, const string& type);
+      const string* tag(int ent);
+      const string* type(int ent);
+      store_entry* get_entry(int ent);
+      void delete_entry(int ent);
```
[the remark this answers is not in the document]
jlmucb: done

include/certifier_framework.h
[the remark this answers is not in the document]
jlmucb: I want a seperate insert

src/certifier.cc
```diff
@@ -25,444 +25,218 @@ using namespace certifier::utilities;
...
-    t->mutable_auth_key()->CopyFrom(psm.channel_authentication_keys(i).auth_key());
-    tkm_[i] = t;
-  }
-  num_blobs_ = psm.blobs_size();
-  for (int i = 0; i < num_blobs_; i++) {
-    tagged_blob_message* t = new(tagged_blob_message);
-    t->CopyFrom(psm.blobs(i));
-    tagged_blob_[i] = t;
+  void certifier::framework::policy_store::delete_entry(unsigned ent) {
+    if (num_ents_ <= 0)
```
gapisback, line 162: Firstly, the check for `<= 0` is unnecessary as `ent` is unsigned, so the value can never be < 0.

Secondly, if num_ents_ currently is 0, `delete_entry(0)` should be a no-op but not really an error.

I think it will be better to change this check to something like:

```
if ((num_ents_ == 0) && (ent == 0))
   return true;   // treat no-op, soft-failure as true
   
// Cannot delete an entry beyond # of entries tracked.
if (ent >= num_ents_)
  return false;
```

src/certifier.cc
```diff
@@ -25,444 +25,218 @@ using namespace certifier::utilities;
...
+      return;
+    delete entry_[ent];
```
gapisback, line 164: Note, if checks suggested on prev line are not right, this will end up deleting some random out-of-bounds entry.

The L165 is also fraught with bad behaviour if say num_ents_ is 5 and `ent` is 10.

And, even worse, the code below will try shift entries up one cell but may never end the loop.

gapisback (changes requested):
HI, @jlmucb -- 

Thanks for addressing some of the comments in my 1st set of comments.

Here are my 2nd round of comments / feedback. 

I went thru your updated change-set, and took a re-look at `certifier_framework.h`, `cc_helpers.cc` and `certifier.c`.

I still see some items not addressed; not sure if you haven't got to them, yet, or we have differing opinions. (Let's talk if that would be faster on items that I have marked but not addressed, yet.)

Will finish the balance of diffs in tests and support.cc after this goes out.

src/store_tests.cc
```diff
@@ -410,3 +217,162 @@ bool test_init_and_recover_containers(bool print_all) {
...
+  }
+
+  if (ps.get_num_entries() != 3) {
+    printf("Error: policy-key store number of entries is %d should be 3\n", ps.get_num_entries());
+    return false;
+  }
+
+  tag1 = "test-entry-2";
+  type1 = "binary";
+  int ent = ps.find_entry_by_tag(tag1);
```
gapisback, line 301: The business of uniqueness on either `(tag)` or on the pair `(tag, type)`, still bothers me.

What if someone adds another entry with same tag: 'test-entry-2'. Then this find() can never return the 2nd entry, till the 1st entry is deleted.

I think this make the interface a bit clunky and under-specified in its usefulness.

Still waiting to hear how you plan to reconcile these uniqueness semantics.

src/store_tests.cc
```diff
@@ -410,3 +217,162 @@ bool test_init_and_recover_containers(bool print_all) {
   return true;
 }
 
+bool test_policy_store(bool print_all) {
```
gapisback, line 220: I think this single large test case handling all sub-ops is difficult to follow and to maintain.

It's ok if you wish to go in like as-is, ... but I think we should talk and I am happy to refactor this into smaller unit-test cases, each for one single operation and its api.

For instance the new code from L352 onwards, `ps.Serialize(&saved))` ... can all go into a sub-case `test_policy_store_serialize_deserialize()` function.

gapisback (commented):
@jlmucb -- This is my 3rd set of comments on the test code file. nothing major here ... some minor remarks.

When you think you've updated the change set addressing all my previous comments, let me know. I would like to take a re-look.

Thanks, for this work!

src/certifier.cc
```diff
@@ -148,18 +140,15 @@ bool certifier::framework::policy_store::put(unsigned ent, const string v) {
...
-      if (!add_entry(tag, type, value)) {
-        return false;
-      } else {
-        return true;
-      }
+      return add_entry(tag, type, value);
     }
-    put(ent, value);
+    if (!put(ent, value))
+      return false;
```
gapisback, line 165: Small request: Like it was done on L143, it will be simpler to replace this if()-else() with a simple: 

`return put(ent, value);`

include/certifier_framework.h
```diff
@@ -63,14 +63,15 @@ namespace certifier {
...
       policy_store();
       ~policy_store();
 
+private:
+      bool add_entry(const string& tag, const string& type, const string& value);
+public:
       bool is_policy_key_valid();
       unsigned get_num_entries();
-      bool add_entry(const string& tag, const string& type, const string& value);
-      int find_entry_by_tag(const string& tag);
```
gapisback: Now that `find_entry_by_tag()` is withdrawn, can we consider simplifying the remaining interface to just:

`find_entry(const string& tag, const string& type)`

The uniqueness is now being enforced by (tag, type), so I feel it's unnecessary to have that longer name.

If ever we find a need to add a new interface with just (tag) or just (type), we can name those methods using the "_by_" sub-term.

gapisback (approved):
Hi, @jlmucb -- THANK YOU, for patiently working through all my comments and for revising the interfaces and implementation. The change-set is looking a lot more solid now.

I've given a few minor requests with this review.

Additionally, I think we should revisit and discuss the following points (in our forthcoming meeting), so as to improve the interfaces and code going forward:

1. Class' definition has multiple `public` sections. Need to re-arrange the class' structure layout  to consolidate these. (Can be done in a follow-on for all APIs rework.)
2. Use of `private` member fields. (Making `add_entry()` private is a good start.)
3. Change return type of `find_entry_by_tag_and_type` to `unsigned`, returning the index of the entry. This allows one to just call: `get_entry(find_entry_by_tag_and_type())`. We can return `(max_num_ents_ + 1)` [ or just `max_num_ents_`] to signal an entry-not-found condition.
4. Change `delete_entry()` to `bool`. Otherwise, error checking is not possible. This also makes all the return from all "do-something" verb-APIs consistently to be `bool`.
5. Separate out newly-added concurrency support to a follow-on PR, along with its own multi-user concurrency tests.

Thanks,
--A>

include/certifier_framework.h
gapisback: @jlmucb -- Btw, I should remark here that an interface with both `tag` and `type` of string type is likely to cause errors, when someone mistakenly interchanges the two arguments.

I don't have any clear suggestion how-to address this potential hole, other than that we may want to consider changing `type` to an int or some enumerated value.

gapisback:
**Summary notes** from final meeting on review - to be done later on:

1. Adding `''` around `'%s'` to be done later for all files.
2. Pull out the synchronization logic to a follow-on change-set. (Look into using a semaphore rather than a mutex.)

PR #152, opened 2023-07-16
Add AMD SEV-SNP VCEK extensions parsing and checks.

yelvmw:
This patch added AMD SEV-SNP VCEK certificate custom extension parsing support. The platform TCB version is now checked as part of the SEV-SNP attestation report verification in addition to signature verification.

### DIFF
--- a/certifier_service/certlib/certlib_proofs.go
+++ b/certifier_service/certlib/certlib_proofs.go
@@ -1475,6 +1475,14 @@ func GetUserDataHashFromSevAttest(binSevAttest []byte) []byte {
 	return []byte(binSevAttest[0x50:0x90])
 }
 
+func GetTcbVersionFromSevAttest(binSevAttest []byte) uint64 {
+	tcb := uint64(0)
+	for i := 0; i < 8; i++ {
+		tcb |= uint64(binSevAttest[0x180+i]) << (8 * i)
+	}
+	return tcb
+}
+
 /*
 	Policy byte:
 		Bit 3: Guest can be activated on multiple sockets.
@@ -1536,14 +1544,7 @@ func GetPlatformFromSevAttest(binSevAttest []byte) *certprotos.Platform {
 	p5 := MakeProperty(pn5, ivt, nil, &ce, &m2iv)
 	props.Props = append(props.Props, p5)
 
-	tcb := uint64(binSevAttest[0x180])
-	tcb = (uint64(binSevAttest[0x181]) << 8) | tcb
-	tcb = (uint64(binSevAttest[0x182]) << 16) | tcb
-	tcb = (uint64(binSevAttest[0x183]) << 24) | tcb
-	tcb = (uint64(binSevAttest[0x184]) << 32) | tcb
-	tcb = (uint64(binSevAttest[0x185]) << 40) | tcb
-	tcb = (uint64(binSevAttest[0x186]) << 48) | tcb
-	tcb = (uint64(binSevAttest[0x187]) << 56) | tcb
+	tcb := GetTcbVersionFromSevAttest(binSevAttest)
 
 	// DEBUG
 	fmt.Printf("tcb: %08x\n", tcb)
@@ -1683,6 +1684,30 @@ func VerifySevAttestation(serialized []byte, k *certprotos.KeyMessage) []byte {
 	if !ecdsa.Verify(PK, hashOfHeader[0:48], r, s) {
 		fmt.Printf("VerifySevAttestation: ecdsa.Verify failed\n")
 		return nil
+	}
+
+	vcekTcbVer := k.GetSnpTcbVersion()
+	// TODO: Test whether the VCEK TCB Version is valid for now.
+	// This will be removed in the future after the simulator change.
+	if vcekTcbVer != ^uint64(0) {
+		tcbVer := GetTcbVersionFromSevAttest(ptr)
+		if vcekTcbVer != tcbVer {
+			fmt.Printf("VerifySevAttestation: Platform TCB Version check failed\n")
+			fmt.Printf("VCEK TCB Version: %08x\n", vcekTcbVer)
+			fmt.Printf("Platform TCB Version: %08x\n", tcbVer)
+			return nil
+		}
+		chipid := ptr[0x1A0:0x1E0]
+		if !bytes.Equal(chipid, k.GetSnpChipid()) {
+			fmt.Printf("VerifySevAttestation: Chipid check failed\n")
+			fmt.Printf("VCEK HwID: ")
+			PrintBytes(k.GetSnpChipid())
+			fmt.Printf("\n")
+			fmt.Printf("Platform Chip ID: ")
+			PrintBytes(chipid)
+			fmt.Printf("\n")
+			return nil
+		}
 	}
 
 	// return measurement, if successful
--- a/certifier_service/certlib/certlib_support.go
+++ b/certifier_service/certlib/certlib_support.go
@@ -1759,6 +1759,18 @@ func GetSubjectNameFromCert(cert *x509.Certificate) *string {
 	return &cert.Subject.CommonName
 }
 
+func GetVcekExtValue(ext pkix.Extension) (uint8, error) {
+	if ext.Value[0] != 0x2 {
+		fmt.Printf("Invalid extension type!\n")
+		return 0, errors.New("Invalid extension type")
+	}
+	if ext.Value[1] != 0x1 && ext.Value[1] != 0x2 {
+		fmt.Printf("Invalid extension length!\n")
+		return 0, errors.New("Invalid extension length")
+	}
+	return ext.Value[ext.Value[1]+1], nil
+}
+
 func GetSubjectKey(cert *x509.Certificate) *certprotos.KeyMessage {
 	name := GetSubjectNameFromCert(cert)
 	if name == nil {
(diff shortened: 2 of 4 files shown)


certifier_service/certlib/certlib_proofs.go
```diff
@@ -1685,6 +1685,24 @@ func VerifySevAttestation(serialized []byte, k *certprotos.KeyMessage) []byte {
...
+	vcekTcbVer := k.GetSnpTcbVersion()
+	if vcekTcbVer != ^uint64(0) {
+		tcbVer := uint64(ptr[0x180])
+		tcbVer = (uint64(ptr[0x181]) << 8) | tcbVer
+		tcbVer = (uint64(ptr[0x182]) << 16) | tcbVer
+		tcbVer = (uint64(ptr[0x183]) << 24) | tcbVer
+		tcbVer = (uint64(ptr[0x184]) << 32) | tcbVer
+		tcbVer = (uint64(ptr[0x185]) << 40) | tcbVer
+		tcbVer = (uint64(ptr[0x186]) << 48) | tcbVer
+		tcbVer = (uint64(ptr[0x187]) << 56) | tcbVer
```
larrydewey, line 1697: This seems like a lot of manual manipulation. Would it make more sense to use a loop here?

certifier_service/certlib/certlib_support.go
```diff
@@ -1781,8 +1793,41 @@ func GetSubjectKey(cert *x509.Certificate) *certprotos.KeyMessage {
 			fmt.Printf("GetSubjectKey: Can't internal ecc public key\n")
 			return nil
 		}
+
+		// Look for AMD VCEK cert extensions if they exist
+		oidBlSPL := "1.3.6.1.4.1.3704.1.3.1"
+		oidTeeSPL := "1.3.6.1.4.1.3704.1.3.2"
+		oidSnpSPL := "1.3.6.1.4.1.3704.1.3.3"
+		oidUcodeSPL := "1.3.6.1.4.1.3704.1.3.8"
```
larrydewey, line 1801:
```suggestion
		oidBlSPL := "1.3.6.1.4.1.3704.1.3.1"
		oidTeeSPL := "1.3.6.1.4.1.3704.1.3.2"
		oidSnpSPL := "1.3.6.1.4.1.3704.1.3.3"
		oidUcodeSPL := "1.3.6.1.4.1.3704.1.3.8"
		oidHwId := "1.3.6.1.4.1.3704.1.4"
```

Looks like this is missing the Hardware ID, which is a critical part of the extension, as it matches the physical chip id of the secure processor belonging to a given node. You will likely want to add a slice comparison function/method to perform this comparison, as it is an array of bytes.

certifier_service/certprotos/certifier.proto
```diff
@@ -73,6 +73,8 @@ message key_message {
   optional bytes other_key_formats          = 8;
   optional string not_before                = 9;
   optional string not_after                 = 10;
+  // Optional SEV-SNP VCEK specific fields
+  optional uint64 snp_tcb_version           = 11;
```
larrydewey, line 77: Would we want a separate field for the chip id here, as well?

src/certifier.cc
```diff
@@ -481,6 +481,74 @@ bool GetX509FromCert(const string& cert, X509* x) {
...
+#define EXT_PRODUCT_NAME    "1.3.6.1.4.1.3704.1.2"
+#define EXT_BLSPL           "1.3.6.1.4.1.3704.1.3.1"
+#define EXT_TEESPL          "1.3.6.1.4.1.3704.1.3.2"
+#define EXT_SNPSPL          "1.3.6.1.4.1.3704.1.3.3"
+#define EXT_SPL4            "1.3.6.1.4.1.3704.1.3.4"
+#define EXT_SPL5            "1.3.6.1.4.1.3704.1.3.5"
+#define EXT_SPL6            "1.3.6.1.4.1.3704.1.3.6"
+#define EXT_SPL7            "1.3.6.1.4.1.3704.1.3.7"
+#define EXT_UCODESPL        "1.3.6.1.4.1.3704.1.3.8"
+#define EXT_HWID            "1.3.6.1.4.1.3704.1.4"
```
larrydewey, line 496: It looks like the verification for this is missing here, as well.

larrydewey (changes requested):

certifier_service/certlib/certlib_support.go
```diff
@@ -1781,8 +1793,41 @@ func GetSubjectKey(cert *x509.Certificate) *certprotos.KeyMessage {
...
+			} else if strings.Contains(ext.Id.String(), oidSnpSPL) {
+				snpSPL = GetVcekExtValue(ext)
+				snpExtExist = true
+			} else if strings.Contains(ext.Id.String(), oidUcodeSPL) {
+				ucodeSPL = GetVcekExtValue(ext)
+				snpExtExist = true
+			}
+		}
+		var tcbVer uint64
+		if snpExtExist {
```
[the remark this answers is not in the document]
yelvmw: Did you mean GetVcekExtValue(ext) could return ^uint8(0) for one of the SPLs? I thought about whether to assert that in the if branches above. If any of the GetVcekExtValue(ext) returned failure, or any of the extensions we want are missing, I guess we can assume the VCEK cert was corrupted and the verification should fail?

certifier_service/certlib/certlib_proofs.go
yelvmw: Will do.

certifier_service/certlib/certlib_support.go
yelvmw: Thanks for the info! I didn't realize it's the chip id in the report. Will add that.

certifier_service/certprotos/certifier.proto
yelvmw: Will do

src/certifier.cc
yelvmw: Will update.

certifier_service/certlib/certlib_support.go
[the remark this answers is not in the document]
larrydewey: Maybe create a type like this, and set the fields to `Spl`?

```Go
type Spl struct {
	field uint8
	isSet bool
}

func (s *Spl) SetField(val uint8) {
	s.field = val
	s.isSet = true
}

func (s Spl) GetField() uint8 {
	return s.field
}

func (s Spl) IsSet() bool {
	return s.isSet
}
```

Then do this:

```suggestion
		if snpExtExist && blSPL.IsSet() && teeSPL.IsSet() && snpSPL.IsSet() && ucodeSPL.IsSet() {
		...
```

certifier_service/certlib/certlib_support.go
```diff
@@ -1759,6 +1759,18 @@ func GetSubjectNameFromCert(cert *x509.Certificate) *string {
 	return &cert.Subject.CommonName
 }
 
+func GetVcekExtValue(ext pkix.Extension) uint8 {
+	if ext.Value[0] != 0x2 {
+		fmt.Printf("Invalid extension type!\n")
+		return ^uint8(0)
```
larrydewey, line 1765: You are returning a `255` here? Is this what you are expecting? Would it make more sense to to an error return here?

certifier_service/certlib/certlib_support.go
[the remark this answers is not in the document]
yelvmw: I could do this, or equivalently, check for whether GetVcekExtValue(ext) returns 0xFF for any of the if statements and bail if any of them does. In this case, tcbVer will be set to 0xFFFFFFFF, which will skip the check (we currently skip the check if the extensions do not exist in the VCEK cert. This makes the simulator cert and attestation verification pass).

Alternatively, do you feel we should fail the verification instead of skipping it in this case?

certifier_service/certlib/certlib_support.go
yelvmw: I'm treating 0xFF as an error here explicitly. If 0xFF is valid for these SPL values, I'll have to change the design.

larrydewey: We may have to change the design here. Even though the field is signed (ASN.1 standard), they zero pad the field, so the overall integer number could still be a max unsigned eight bit integer.

yelvmw: Understood. Thanks for the info. I'll update both the C++ and Golang code for that.

certifier_service/certlib/certlib_support.go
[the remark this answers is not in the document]
larrydewey: > Did you mean GetVcekExtValue(ext) could return ^uint8(0) for one of the SPLs?

I was thinking more along the lines of if the extension isn't found, the default value in Go is zero, which is technically a valid value.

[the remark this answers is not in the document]
yelvmw: Understood. My current thought is like this:

1. All values from 0x0 - 0xFF are valid extension values for the various SPLs
2. GetVcekExtValue(ext) will return {bool res, uint8 value} pair
3. If any of the extensions are missing, we will bail and set tcbVer to 0xFFFFFFFF
4. If tcbVer is 0xFFFFFFFF, then tcb version check during the report verification will be skipped

Now, the questions are: 1. Is bullet 3 assumption reasonable? And 2. Is bullet 4 logic reasonable?

Additionally, I do also have a question about ucodeSPL. In the cert, it seems to be a 16 bit value. But when the specification talks about TCB_VERSION, it assumes ucodeSPL is 8 bit. In the code above, I assumed ucodeSPL is 8 bit instead. Is this OK?

src/certifier.cc
```diff
@@ -246,6 +246,101 @@ bool GetX509FromCert(const string& cert, X509* x) {
...
+  }
+  idx = X509_get_ext_by_NID(vcek, nid, -1);
+  if (idx == -1) {
+    return false;
+  }
+
+  ex = X509_get_ext(vcek, idx);
+  extvalue = X509_EXTENSION_get_data(ex);
+  extlen = ASN1_STRING_length(extvalue);
+  vals = (unsigned char *) ASN1_STRING_get0_data(extvalue);
```
larrydewey, line 283: Is there a reason we are removing the `const` context here from openssl? It doesn't look like we are modifying it in any way. If we are going to cast, would it make more sense to do:

```suggestion
  vals = reinterpret_cast<unsigned char *>(ASN1_STRING_get0_data(extvalue));
```

src/certifier.cc
```diff
@@ -246,6 +246,101 @@ bool GetX509FromCert(const string& cert, X509* x) {
...
```
larrydewey, line 333: Same question here as from [previously asked](https://github.com/vmware-research/certifier-framework-for-confidential-computing/pull/152/files#r1268197716)

src/certifier.cc
```diff
@@ -246,6 +246,101 @@ bool GetX509FromCert(const string& cert, X509* x) {
...
+
+  if (idlen < extlen || chipid == nullptr) {
+    return false;
+  }
+
+  memcpy(chipid, vals, extlen);
```
larrydewey, line 339: Where these are both the same type, would it make sense to:

```c++
#include <algorithm>
```
and then use:

```suggestion
  std::copy(vals, vals + extlen, chipid);
```

yelvmw: Will do.

larrydewey (approved):
Looks good! Thanks for letting me chime in!

yelvmw:
Thank you for the review, Larry!

jlmucb (approved):

PR #159, opened 2023-07-20
Add custom extensions to simulator VCEK certificate.

yelvmw:
Added custom extensions with fixed values to the simulator VCEK certificates we generate. The VCEK certificate extension verification check is also removed so we always check for the extensions during report verification.

### DIFF
--- a/src/support.cc
+++ b/src/support.cc
@@ -27,6 +27,7 @@
 
 #include "support.h"
 #include "certifier.pb.h"
+#include "sev-snp/sev_vcek_ext.h"
 
 #include <sys/ioctl.h>
 #include <sys/socket.h>
@@ -2627,7 +2628,8 @@ bool certifier::utilities::produce_artifact(
                       key_message& signing_key, string& issuer_name_str,
                       string& issuer_organization_str, key_message& subject_key,
                       string& subject_name_str, string& subject_organization_str,
-                      uint64_t sn, double secs_duration, X509* x509, bool is_root) {
+                      uint64_t sn, double secs_duration, X509* x509, bool is_root,
+                      bool vcek /* default =false */) {
 
   ASN1_INTEGER* a = ASN1_INTEGER_new();
   ASN1_INTEGER_set_uint64(a, sn);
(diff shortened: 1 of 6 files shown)


src/support.cc
```diff
@@ -32,6 +32,18 @@
 #include <sys/socket.h>
 #include <string>
 
+#define EXT_STRUCT_VERSION  "1.3.6.1.4.1.3704.1.1"
```
gapisback, line 35: Looking at the usage of these mnemonics, they seem to be only used in the `else` block in `src/support.cc`, here:

```
2673   if (!vcek) {
2674     add_ext(x509, NID_key_usage, "critical,keyCertSign,digitalSignature,cRLSign");
2675     add_ext(x509,  NID_ext_key_usage, "clientAuth,serverAuth");
2676     // add_ext(x509, NID_subject_key_identifier, "hash");
2677     if (is_root) {
2678       add_ext(x509, NID_basic_constraints, "critical,CA:TRUE");
2679     }
2680   } else {
2681     // Add custom extensions if we are generating VCEK for the SEV-SNP simulator
```

So if these constants are all VCEK-related, how about naming them all with `VCEK_` prefix, like `VCEK_STRUCT_VERSION`, `VCEK_BLSPL` and so on.

`EXT` just means external, but is not indicative that these mnemonics are really only related to VCEK-stuff.

If you do want to retain `EXT` also, the preferred name would be `VCEK_EXT_STRUCT_VERSION` and so on. A bit longer ... but less ambiguous.

gapisback: Also, over time, the provenance of these constants will be lost. 

If they are coming from some external AMD-VCEK spec, can you xref its url here?

If they are simply some random constants chosen to exercise the test-case & simulator, can you please add a comment to that effect. Something on the lines of:

"Constants are chosen arbitrarily just to have some data for the custom extensions to check-against in the simulated VCEK-certificates we internally generate for testing."

gapisback (approved):
Took a quick look; some minor suggestions.

In a past meeting, we had talked about possibly checking-in to the repo some Certs to do this kind of testing. Is this change-set related to that topic? If so, do we [not] expect to see some cert-files being checked-in? ... May be I'm mistaken.

With this change, I am assuming that these changes are exercised / tested in this chunk in build.yml:

```
test-sev-snp-simulator-sev-test
sudo ./sev-test
```

If not, please somehow ensure that this new code of adding dummy values to the generated VCEK-certificates is always being exercised in CI, thru some test case.

src/support.cc
yelvmw: These are OIDs applied by and belong to AMD. So, I will add the xref of the spec. Will also update the macro names.

yelvmw:
Updated according to Aditya's comments. After this patch, the VCEK extension code path is exercised under CI all the time as a result of the SEV-SNP simulator exposure.

Issue #170, opened 2023-08-04
Exercising default constructor of cc_trust_data() runs into a seg-fauilt.


gapisback:
The following seg-fault was discovered while working on PR #158 to add new Python tests that exercise class / methods added recently under PR #162 (@ SHA a167530).

A simplest pytest to instantiate a `cc_trust_data()` object with a default constructor runs into the following stack:

```
 57 def test_cf_cc_trust_data_default_ctor():
 58     """
 59     Basic exerciser for an empty cc_trust_data() object.
 60     """
 61     cctd = libcf.new_cc_trust_data()
 62     libcf.delete_cc_trust_data(cctd)
```

Stack is:

```
cc-sdb-vm:[81] $ PYTHONPATH=../.. pytest --capture=tee-sys -v test_cert_framework_basic.py -k test_cf_cc_trust_data
=================================================== test session starts ====================================================
platform linux -- Python 3.10.12, pytest-7.4.0, pluggy-1.2.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /home/agurajada/Code/certifier-framework-for-confidential-computing/tests/pytests
collected 7 items / 5 deselected / 2 selected

test_cert_framework_basic.py::test_cf_cc_trust_data_default_ctor Fatal Python error: Segmentation fault

Current thread 0x00007f6ab754a000 (most recent call first):
  File "/home/agurajada/Code/certifier-framework-for-confidential-computing/tests/pytests/test_cert_framework_basic.py", line 61 in test_cf_cc_trust_data_default_ctor
  File "/home/agurajada/.local/lib/python3.10/site-packages/_pytest/python.py", line 194 in pytest_pyfunc_call
  File "/home/agurajada/.local/lib/python3.10/site-packages/pluggy/_callers.py", line 80 in _multicall
  File "/home/agurajada/.local/lib/python3.10/site-packages/pluggy/_manager.py", line 112 in _hookexec
  File "/home/agurajada/.local/lib/python3.10/site-packages/pluggy/_hooks.py", line 433 in __call__
  File "/home/agurajada/.local/lib/python3.10/site-packages/_pytest/python.py", line 1788 in runtest
  File "/home/agurajada/.local/lib/python3.10/site-packages/_pytest/runner.py", line 169 in pytest_runtest_call
  File "/home/agurajada/.local/lib/python3.10/site-packages/pluggy/_callers.py", line 80 in _multicall
[...]

PR #171, opened 2023-08-04
(#170) Refactor cc_trust_data() constructor to fix bug.

gapisback:
Default constructor for cc_trust_data() ran into seg-faults due to use of uninitialized variables. Refactor the constructor to add  a common private `cc_trust_data_default_init()` method, shared by both constructors. 

Add Python tests which trigger this bug and test the fix.

Good catch!

The only concern I have is that the destructor does not delete the certifiers structures that were newed when adding.

You are correct that the code that does that, below, should be in the destructor (Man, that was a wierd mistake).

  for (int i = 0; i < num_certified_domains_; i++) {
      if (certified_domains_[i] != nullptr) {
        delete certified_domains_[i];
        certified_domains_[i] = nullptr;
      }
  }
  delete certified_domains_;
  certified_domains_ = nullptr;
  num_certified_domains_ = 0;

### DIFF
--- a/src/cc_helpers.cc
+++ b/src/cc_helpers.cc
@@ -117,6 +117,15 @@ void certifier::framework::cc_trust_data::cc_trust_data_default_init() {
 }
 
 certifier::framework::cc_trust_data::~cc_trust_data() {
+  for (int i = 0; i < num_certified_domains_; i++) {
+      if (certified_domains_[i] != nullptr) {
+        delete certified_domains_[i];
+        certified_domains_[i] = nullptr;
+      }
+  }
+  delete certified_domains_;
+  certified_domains_ = nullptr;
+  num_certified_domains_ = 0;
 }
 
 bool certifier::framework::cc_trust_data::cc_all_initialized() {
@@ -1247,7 +1256,7 @@ bool certifier::framework::cc_trust_data::certify_secondary_domain(const string&
     }
   }
 
-  return found->certify_domain();
+  return (found ? found->certify_domain() : false);
 }
 
 bool certifier::framework::cc_trust_data::init_peer_certification_data(const string& public_key_alg) {
--- a/src/cc_helpers.cc
+++ b/src/cc_helpers.cc
@@ -82,16 +82,22 @@ extern string gramine_platform_cert;
 
 certifier::framework::cc_trust_data::cc_trust_data(const string& enclave_type, const string& purpose,
     const string& policy_store_name) {
-
+  cc_trust_data_default_init();
   if (purpose == "authentication" || purpose == "attestation") {
     purpose_= purpose;
     cc_basic_data_initialized_ = true;
-  } else {
-    cc_basic_data_initialized_ = false;
-    purpose_ = "unknown";
   }
   enclave_type_ = enclave_type;
   store_file_name_ = policy_store_name;
+}
+
+certifier::framework::cc_trust_data::cc_trust_data() {
+  cc_trust_data_default_init();
+}
+
+void certifier::framework::cc_trust_data::cc_trust_data_default_init() {
+  cc_basic_data_initialized_ = false;
+  purpose_ = "unknown";
   cc_policy_info_initialized_= false;
   cc_policy_store_initialized_ = false;
   cc_service_key_initialized_ = false;
@@ -108,29 +114,6 @@ certifier::framework::cc_trust_data::cc_trust_data(const string& enclave_type, c
   for (int i = 0; i < max_num_certified_domains_; i++) {
       certified_domains_[i] = nullptr;
   }
-}
-
-certifier::framework::cc_trust_data::cc_trust_data() {
-  cc_basic_data_initialized_ = false;
-  cc_policy_info_initialized_= false;
-  cc_policy_store_initialized_ = false;
-  cc_service_key_initialized_ = false;
-  cc_service_cert_initialized_ = false;
-  cc_service_platform_rule_initialized_ = false;
-  cc_sealing_key_initialized_ = false;
-  cc_provider_provisioned_ = false;
-  x509_policy_cert_ = nullptr;
-  cc_is_certified_ = false;
-
-  for (int i = 0; i < num_certified_domains_; i++) {
-      if (certified_domains_[i] != nullptr) {
-        delete certified_domains_[i];
-        certified_domains_[i] = nullptr;
-      }
-  }
-  delete certified_domains_;
-  certified_domains_ = nullptr;
-  num_certified_domains_ = 0;
 }
 
 certifier::framework::cc_trust_data::~cc_trust_data() {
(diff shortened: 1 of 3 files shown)


src/cc_helpers.cc
```diff
@@ -82,16 +82,22 @@ extern string gramine_platform_cert;
 
 certifier::framework::cc_trust_data::cc_trust_data(const string& enclave_type, const string& purpose,
     const string& policy_store_name) {
-
+  cc_trust_data_default_init();
```
gapisback, line 85: Rearrange the constructor to do the "default initialization" first, followed by the init's needed by passed-in args.

src/cc_helpers.cc
```diff
@@ -82,16 +82,22 @@ extern string gramine_platform_cert;
...
-  } else {
-    cc_basic_data_initialized_ = false;
-    purpose_ = "unknown";
   }
   enclave_type_ = enclave_type;
   store_file_name_ = policy_store_name;
+}
+
+certifier::framework::cc_trust_data::cc_trust_data() {
+  cc_trust_data_default_init();
```
gapisback, line 95: Default constructor will now call default initialization routine.

src/cc_helpers.cc
```diff
@@ -110,29 +116,6 @@ certifier::framework::cc_trust_data::cc_trust_data(const string& enclave_type, c
...
-  cc_policy_store_initialized_ = false;
-  cc_service_key_initialized_ = false;
-  cc_service_cert_initialized_ = false;
-  cc_service_platform_rule_initialized_ = false;
-  cc_sealing_key_initialized_ = false;
-  cc_provider_provisioned_ = false;
-  x509_policy_cert_ = nullptr;
-  cc_is_certified_ = false;
-
-  for (int i = 0; i < num_certified_domains_; i++) {
```
gapisback, line 125: In the old code, this variable `num_certified_domains_` is used uninitialized.
That's causing a seg-fault.

src/cc_helpers.cc
```diff
@@ -110,29 +116,6 @@ certifier::framework::cc_trust_data::cc_trust_data(const string& enclave_type, c
...
-      if (certified_domains_[i] != nullptr) {
-        delete certified_domains_[i];
-        certified_domains_[i] = nullptr;
-      }
-  }
-  delete certified_domains_;
```
gapisback, line 131: This and delete on L127 seem like destructor code.

To me, seems like this old code was mixing-up logic / code for constructor and destructor.

tests/pytests/test_cert_framework_basic.py
```diff
@@ -53,6 +53,29 @@ def test_getmembers_of_cc_framework():
     for item in ccf_builtin_methods:
         print(' -', item[0], item[1])
 
+# ##############################################################################
+def test_cf_cc_trust_data_default_ctor():
+    """
+    Basic exerciser for an empty cc_trust_data() object.
+    """
+    cctd = libcf.new_cc_trust_data()
```
gapisback, line 61: Without the c++ fix, this single line will trip the seg-fault consistently on every new run.

src/cc_helpers.cc
jlmucb: OK

src/cc_helpers.cc
jlmucb: great!

jlmucb (changes requested):
Good catch!

The only concern I have is that the destructor does not delete the certifiers structures that were newed when adding.

You are correct that the code that does that, below, should be in the destructor (Man, that was a wierd mistake).

for (int i = 0; i < num_certified_domains_; i++) {
  if (certified_domains_[i] != nullptr) {
    delete certified_domains_[i];
    certified_domains_[i] = nullptr;
  }
}
delete certified_domains_;
certified_domains_ = nullptr;
num_certified_domains_ = 0;
------------------------------------

If you agree and fix that, I'm good with this.  I'm glad you found the problem. Good work.

src/cc_helpers.cc
```diff
@@ -117,6 +117,15 @@ void certifier::framework::cc_trust_data::cc_trust_data_default_init() {
 }
 
 certifier::framework::cc_trust_data::~cc_trust_data() {
+  for (int i = 0; i < num_certified_domains_; i++) {
```
gapisback, line 120: @jlmucb -- Restored the correct delete / release-memory code in the desctructor.

See my notes in the test case.

src/cc_helpers.cc
```diff
@@ -1247,7 +1256,7 @@ bool certifier::framework::cc_trust_data::certify_secondary_domain(const string&
     }
   }
 
-  return found->certify_domain();
+  return (found ? found->certify_domain() : false);
```
gapisback, line 1259: This fixes another seg-fault in this method. If supplied domain-name is not found, `found` will be NULL.

Handle that.

tests/pytests/test_cert_framework_basic.py
```diff
@@ -76,6 +76,45 @@ def test_cf_cc_trust_data():
...
+
+    result = libcf.cc_trust_data_add_or_update_new_domain(cctd,
+                                                          'test-security-domain',
+                                                          'test-dummy-certificate',
+                                                          'localhost', 8121,
+                                                          'localhost', 8123)
+    assert result is True
+
+    libcf.delete_cc_trust_data(cctd)
```
gapisback, line 95: This delete call in new test case will invoke the `cc_trust_data()` destructor.

I didn't quite find a way to confirm or check thru pytests that there is a memory leak when the new C++ code in the destructor was not added.

But from code inspection, the memory release code in the desctructor is fine, so I'm ok with just this exerciser unit test.

tests/pytests/test_cert_framework_basic.py
```diff
@@ -76,6 +76,45 @@ def test_cf_cc_trust_data():
...
+    cctd = libcf.new_cc_trust_data()
+
+    result = libcf.cc_trust_data_add_or_update_new_domain(cctd,
+                                                          'test-security-domain',
+                                                          'test-dummy-certificate',
+                                                          'localhost', 8121,
+                                                          'localhost', 8123)
+    assert result is True
+
+    result = libcf.cc_trust_data_certify_secondary_domain(cctd,
```
gapisback, line 112: This lines triggers, and verifies the fix for, the seg-fault in this `certify_secondary_domain()` method.

gapisback:
@jlmucb -- I have addressed your comment to add the code in the destructor.

Also, found and fixed another seg-fault. Plus added 2 new tests.

ready for re-review.

I didn't run the multi-domain app exercise workflow ... as it's too time consuming. I think the changes are fine based on inspection.

jlmucb (approved):
Good catch!

The only concern I have is that the destructor does not delete the certifiers structures that were newed when adding.

You are correct that the code that does that, below, should be in the destructor (Man, that was a wierd mistake).

for (int i = 0; i < num_certified_domains_; i++) {
  if (certified_domains_[i] != nullptr) {
    delete certified_domains_[i];
    certified_domains_[i] = nullptr;
  }
}
delete certified_domains_;
certified_domains_ = nullptr;
num_certified_domains_ = 0;
------------------------------------

If you agree and fix that, I'm good with this.  I'm glad you found the problem. Good work.

Issue #181, opened 2023-08-19
Memory leak detected by pytests invoking init_client_ssl() interface: memory leak of type 'key_message *', no destructor found.


gapisback:
While developing new pytests to invoke Python bindings to Certifier Framework interfaces, the following memory leak was detected and reported by the pytests machinery:

```
test_certifier_framework.py::test_run_app_as_a_client_init_client_ssl PublicKeyFromCert() error, line 432, Failed to retrieve HWID from VCEK extensions
 ... cctd.init_policy_key() succeeded.
 ... cctd.initialize_simulated_enclave_data() succeeded.
 ... cctd.cold_init() succeeded.
 ... cctd.warm_restart() succeeded.
certify_domain():1539: enclave_type_ = 'simulated-enclave', purpose_ = 'authentication'
 ... cctd.certify_me() succeeded.
 ... cctd.cc_all_initialized() succeeded.
 ... Secure channel client instantiated.
 ... cctd.trust data is initialized.
 ... cctd.primary admissions cert is valid.
asn1_to_x509() error, line: 3764, no x509 pointer
init_client_ssl() error, line 2409, init_client_ssl: root cert invalid
64756d6d792d73657269616c697a65642d706f6c6963792d63657274
swig/python detected a memory leak of type 'key_message *', no destructor found.
```

The test case itself is currently not working due to some Python arg-calling interface issues. 
After correcting those manually (by patching generated SWIG wrapper code), the memory leak messages are still seen:

**swig/python detected a memory leak of type 'key_message *', no destructor found.**

--
The suspected code is probably here in the destructor method of `secure_authenticated_channel()`:

```
2362 certifier::framework::secure_authenticated_channel::
2363     ~secure_authenticated_channel() {
2364   role_.clear();
2365   channel_initialized_ = false;
2366
2367   // ? FIXME - Seems to cause a memory leak detected in pytests
2368   // delete private_key_
2369
2370   if (ssl_ctx_ != nullptr)
```
Investigate this code flow to see how to do a `delete` on protobuf-generated sub-objects hanging off of `secure_authenticated_channel()`.

PR #182, opened 2023-08-19
Add SWIG-pytests and interface file to verify typemap'ping rules.

gapisback:
This commit adds stripped down versions of certifier_framework.h
interfaces in new files: swigpytest.h, swigpytest.cc with the driving
SWIG swigpytest.i interfaces file. Makefile rules are added to build
the libswigpytest.so shared library. New pytests are added
to exercise SWIG-Python interface mapping rules.

Minor adjustments are introduced in Certifier interfaces to enable Python bindings.

New target 'swigpytestsharedlib' added to certifier.mak, to create
the shared library and Python maping files to exercise newly added tests in test_libswigpytest.py .

----
NOTE: TO reviewers:

This change-set basically brings-in swigpytest.i interfaces file as a way to unit-test our understanding of SWIG type mapping rules, by exercising mock interfaces added in `swigpytest.h`, with mock implementations added in `swigpytest.cc`.

Small adjustments are done to existing Certifier interfaces to add `const` in places where string args were only used as input.

The things you should review is how this SWIG-pytest interfaces and pytests are coming together to externalize the SWIG interface mappings. Largely, consider this as testing-only change-set, purely to externalize and bring forward a large body "understanding" work I've been engaged in for past 1-2 weeks. 

These pytests are building blocks which provides a framework to build more SWIG interface rules to handle, for example, the invocation of Python methods to pass-down certificates that contain Unicode chars (a blocking bug in current machinery for us).


.github/workflows/build.yml
```diff
@@ -64,6 +64,7 @@ jobs:
         # Rebuild special-target which will invoke -fPIC flag during compilation.
         set -x
         make -f certifier.mak --always-make -j2 sharedlib
+        make -f certifier.mak --always-make -j2 swigpytestsharedlib
```
gapisback, line 67: Invoke make for new target to build SWIG-pytests shared library.

.github/workflows/build.yml
```diff
@@ -120,15 +121,23 @@ jobs:
 
         PYTHONUNBUFFERED=TRUE pytest -v test_libcertifier_framework.py
 
-        PYTHONUNBUFFERED=TRUE pytest -v test_certifier_framework.py
+        # NOTE: Some test cases in test_certifier_framework.py need the Certifier Service
```
gapisback, line 124: The comments should be self-explanatory as to why I had to do this jugglery in build.xml workflow steps.

include/certifier_framework.h
```diff
@@ -373,7 +374,7 @@ class secure_authenticated_channel {
 
 bool server_dispatch(const string &host_name,
                      int           port,
-                     string &      asn1_root_cert,
+                     const string &asn1_root_cert,
```
gapisback, line 377: Same reason here, too ... Input `string` is never changed.

include/certifier_framework.i
```diff
@@ -18,10 +18,11 @@
 %pybuffer_binary(byte* asn1_cert, int asn1_cert_size)   // cc_trust_data()->init_policy_key()
 
 // Xform interfaces returning a string output param to return string * <function>
-%apply string * OUTPUT { string *v }             // policy_store()->get()
-%apply string * OUTPUT { string *psout }         // policy_store()->Serialize()
-%apply string * INPUT  { string& role};          // secure_authenticated_channel() constructor
```
gapisback, line 23: By testing , I think I have now discovered that these typemap rules are not needed if we are simply passing-in string as input arg.

include/swigpytest.h
```diff
@@ -0,0 +1,105 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// ****************************************************************************
+// These are just dummy class definitions, based on the stuff defined in
+// certifier_framework.h . These definitions and interfaces are used solely:
+// - to exercise different combinations of SWIG-interface rules
+// - to verify via pytests the way to exercise these interfaces
+// - to standardize on code structure to exchange data from Python code to C++
```
gapisback, line 21: This comment block should help you understand why I am adding this test-only .h file.

We need to discuss the "right" place for this .h file. Currently, it's simple for the Make-rules to find this in the `include/` dir, but that may not be the ideal place.

I don't want to "fix" the right location right now ... Let's wait for all this bindings work to stabilize and then we can re-work the final dir-location for this file (is my thinking).

src/cc_helpers.cc
```diff
@@ -2359,7 +2363,10 @@ certifier::framework::secure_authenticated_channel::
     ~secure_authenticated_channel() {
   role_.clear();
   channel_initialized_ = false;
-  // delete?
+
+  // ? FIXME - Seems to cause a memory leak detected in pytests
+  // delete private_key_
```
gapisback, line 2368: @jlmucb should fix issue #181

src/cc_helpers.cc
```diff
@@ -2383,24 +2390,26 @@ certifier::framework::secure_authenticated_channel::
...
   OPENSSL_init_ssl(0, NULL);
   SSL_load_error_strings();
 
   private_key_.CopyFrom(private_key);
+
   asn1_root_cert_.assign((char *)asn1_root_cert.data(), asn1_root_cert.size());
+
   root_cert_ = X509_new();
-  if (!asn1_to_x509(asn1_root_cert, root_cert_)) {
+  if (!asn1_to_x509(asn1_root_cert_, root_cert_)) {
```
gapisback, line 2405: I think this change and the change on L2409, L2412 should be ok, as we are copying-in the input `asn1_root_cert` to this member field on L2402, above.

Do give this a good look and confirm that I'm not making a mistake here ...

src/certifier.mak
```diff
@@ -55,22 +55,42 @@ LINK=g++
 # Point this to the right place, if you have to, based on your machine's install:
 # PROTO=/usr/local/bin/protoc
 PROTO=protoc
+
+# protoc --python_out will generate this file
+PROTOC_PB2_PY = certifier_pb2.py
```
gapisback, line 60: many of the changes here are like this one, introducing symbols to refactor repeatedly used names of targets, files etc.

src/certifier.mak
```diff
@@ -55,22 +55,42 @@ LINK=g++
...
 SWIG=swig
 
+# SWIG Debugging flags: Invoke as: $ SWIG_DEBUG=1 make ...
+#   -debug-tmsearch : Show typemap rules searched while processing interface
+#   -debug-tmused   : Show typemap rules finally applied
+#   -E              : Add this to display preprocessed output
+SWIG_DEBUG_FLAGS =
+ifdef SWIG_DEBUG
+SWIG_DEBUG_FLAGS = -debug-tmsearch -debug-tmused -E
+endif
```
gapisback, line 75: I'm just providing this make-hook to invoke as:

```
SWIG_DEBUG=1 make -f certifier.mak ...
```

So that SWIG built-in debugging flags can be turned ON.

These SWIG-debug flags are useful to understand and debug swig typemapping rules / code-flow.

src/swigpytest.cc
```diff
@@ -0,0 +1,137 @@
...
+  role_ = role;
+}
+
+swigpytests::secure_authenticated_channel::~secure_authenticated_channel() {}
+
+bool swigpytests::secure_authenticated_channel::init_client_ssl(
+    const string &asn1_root_cert) {  // In
+
+  asn1_root_cert_ = asn1_root_cert;
+  swig_wrap_fn_name_.assign("init_client_ssl-const-string-asn1_root_cert");
```
gapisback, line 50: This member field provides a testing-hook.

Each method that is invoked thru these SWIG wrappers will store a bread-crumb in this field, indicating which method was actually executed.

The python pytest code will cross-check this and assert the expected method's bread-crumb string.

If ever, due to changes in SWIG typemapping rules, if the flow is changed, the pytest will stumble upon this assert failing, which will help our debugging.

include/certifier_framework.h
jlmucb: yup

jlmucb (approved):

src/cc_helpers.cc
yelvmw: Was this a bug before the change? It seems that d2i_X509 treats ppin as const. Why do we need asn1_root_cert_ to begin with?

gapisback: hi, @yelvmw -- The only diff in this function is the change of the input arg from `string &      asn1_root_cert` to `const string &asn1_root_cert`, as this input argument is read-only.

No, there is no  bug in this code prior to this change. Old code use the input arg whereas the new code is using, effectively, the same param after copying input param to `asn1_root_cert_` member field, on this L :

`2402   asn1_root_cert_.assign((char *)asn1_root_cert.data(), asn1_root_cert.size());`

Your question seems fair. This class member-field `asn1_root_cert_` seems to be used mainly for copying stuff around. 

I dug around and it does seem likely that we could eliminate this field, but I don't want to change unrelated stuff under this change-set. When / if the larger class layout reorgs occur, we should examine and rework the need for such fields.

gapisback:
@ye - I have responded to the single comment you had. If you have other comments / questions, mark them up. I will process those offline in a day or so. Otherwise, can you clear this review so I can do a anity check and prepare for review.

src/swigpytest.cc
yelvmw: Does it make sense to put this file into "tests/pytests"?

yelvmw (approved):
Let's investigate whether it makes sense to move the swigpytest.cc to tests directory. Otherwise LGTM.

src/swigpytest.cc
gapisback: @yelvmw - The 'problem' with that is that this file has to be linked with the generated `wrap*.cc` file. We will need to write a new Makefile in `tests/pytests` to compile these files. Or have the `certifier.mak` reference the dir paths in `tests/pytests` etc.

There are other files such as `certifier_tests.cc` in this `src/` dir, so may be I thought it would be ok for the short-term to leave this file in this dir. However, maybe I should rename it as `swigpytests.cc` for consistency with other `*tests.cc` files.

A related issue is the location of `swigpytest.h` and `swigpytest.i` in include/ dir. That's messy too, and I wasn't too happy with that. 

How about I try to do this:

- Create under `tests/pytests` two dirs: `src/` and `include/`
- Move the .h and .i files to this new `include/` dir.
- Move the `swigpytests.cc` to this new `src/` dir

Add a new Makefile to this area ... and rework the build commands.

If you are agreeable with some approach on these lines, I'll rework stuff. Won't know how things work out till I try.

yelvmw: I see the concern now. I'll leave it to you then. Your plan about src/ and include/ in pytests sounds good to me. But it turned out to be a big hassle, I can live with the current solution too. Thanks!

gapisback: @yelvmw -- I reworked the files in this change set to be `swigpytests.cc`, `swigpytests.h` etc.

The rearrangement of these files to live inside `tests/pytests` will will take time.

Also see my [previous comment](https://github.com/vmware-research/certifier-framework-for-confidential-computing/pull/182#discussion_r1309524416) with my thoughts on this file-location aspect.

Let's wait for this area to settle down ... and then will rework.

Issue #180, opened 2023-08-19
Misleading 'error' messages seen from VCEK-related code flow in 'simulated-enclave' pytests.


gapisback:
Following 'error' messages are seen being emitted while running a collection of new Python pytests being developed as part of SWIG/Python bindings support. (This work is being done testing ground work to address issue #156.)

The messages are as below:
```
test_certifier_framework.py::test_cc_trust_data_get_certified vcek_ext_byte_value() error, line 233, Failed to create NID
get_chipid_from_vcek() error, line 287, Failed to create NID
PublicKeyFromCert() error, line 432, Failed to retrieve HWID from VCEK extensions
 ... cctd.init_policy_key() succeeded.
```

The actual interface works correctly except that these messages are 'misleading'. 

Rework the code to either report benign / informational messages. Or, rework the build rules to not include VCEK-code into shared libraries that will be used for Python testing. 

From an earlier internal email exchange, 

This message “Failed to retrieve …” comes from this block in `certifier.cc`:

404 #ifdef SEV_SNP
405   // If we have VCEK in the policy in the future, this takes care of the extensions.
406   k->set_snp_tcb_version(get_tcb_version_from_vcek(x));
407   memset(chipid, 0, CHIP_ID_SIZE);
408   if (!get_chipid_from_vcek(x, chipid, CHIP_ID_SIZE)) {
409     printf("%s() error, line %d, Failed to retrieve HWID from VCEK extensions\n",
410       __func__, __LINE__);
411   }
412   k->set_snp_chipid(chipid, CHIP_ID_SIZE);
413 #endif

But, this block should only apply for SEV-SNP build on a SEV-enabled platform, right?

The problem seems like from the certifier.mak file which has this at the very top:

  2 #    File: certifier.mak
  3
  4 ENABLE_SEV=1

----
To this, @yelvmw said:

"This failure is benign. The message should not be that scary. I think we should rephrase it: “Failed to retrieve HWID from VCEK extensions. Setting CHIPID to 0.”. As to SEV_SNP def, I think it makes sense to be more conservative if we can."

"I think that’s already the case since ENABLE_SEV is not defined in the app makefiles. If we are building an archive (or shared object) for general consumption, it might make sense to include all platforms. This is an issue for discussion for sure."

gapisback:
These issues have been addressed and integrated to `/main` under PR #183:
```
[...]

PR #183, opened 2023-08-19
(#180) Rework few error msgs & inclusion of SEV_SNP code in shared libs

gapisback:
This commit does minor tweaks to couple of 'error' messages emitted from SEV_SNP-specific VCEK-related code fragments. 

- Deleted one error message, and change another to Info category.

The remaining message will still be raised when running pytests, but it's benign, so should not be troubling to end-users.

### DIFF
--- a/src/certifier.cc
+++ b/src/certifier.cc
@@ -230,7 +230,6 @@ static bool vcek_ext_byte_value(X509 *         vcek,
   // Use OID for both lname and sname so OBJ_create does not fail
   nid = OBJ_create(oid, oid, oid);
   if (nid == NID_undef) {
-    printf("%s() error, line %d, Failed to create NID\n", __func__, __LINE__);
     return false;
   }
   idx = X509_get_ext_by_NID(vcek, nid, -1);
@@ -276,7 +275,7 @@ uint64_t get_tcb_version_from_vcek(X509 *vcek) {
   return tcb_version;
 }
 
-bool get_chipid_from_vcek(X509 *vcek, unsigned char *chipid, int idlen) {
+static bool get_chipid_from_vcek(X509 *vcek, unsigned char *chipid, int idlen) {
   int                  nid = -1, idx = -1, extlen = -1;
   X509_EXTENSION *     ex = NULL;
   ASN1_STRING *        extvalue = NULL;
@@ -284,7 +283,7 @@ bool get_chipid_from_vcek(X509 *vcek, unsigned char *chipid, int idlen) {
 
   nid = OBJ_create(VCEK_EXT_HWID, VCEK_EXT_HWID, VCEK_EXT_HWID);
   if (nid == NID_undef) {
-    printf("%s() error, line %d, Failed to create NID\n", __func__, __LINE__);
+    printf("%s() Warning, line %d, Failed to create NID\n", __func__, __LINE__);
     return false;
   }
   idx = X509_get_ext_by_NID(vcek, nid, -1);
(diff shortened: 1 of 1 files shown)


src/certifier.cc
```diff
@@ -276,15 +276,15 @@ uint64_t get_tcb_version_from_vcek(X509 *vcek) {
...
+static bool get_chipid_from_vcek(X509 *vcek, unsigned char *chipid, int idlen) {
   int                  nid = -1, idx = -1, extlen = -1;
   X509_EXTENSION *     ex = NULL;
   ASN1_STRING *        extvalue = NULL;
   const unsigned char *vals = NULL;
 
   nid = OBJ_create(VCEK_EXT_HWID, VCEK_EXT_HWID, VCEK_EXT_HWID);
   if (nid == NID_undef) {
-    printf("%s() error, line %d, Failed to create NID\n", __func__, __LINE__);
+    printf("%s() Warning, line %d, Failed to create NID\n", __func__, __LINE__);
```
gapisback, line 287: We don't have a proper protocol for message severity levels. 

For now, as this message isn't really an 'error' in all cases, and as the single-caller seems to cope with a `return false`, I changed this from `error` to `Warning`.

@yelvmw -- Please advise, if this is ok, or should I just leave this as `error`.

src/certifier.cc
```diff
@@ -426,13 +426,13 @@ bool PublicKeyFromCert(const string &cert, key_message *k) {
   k->set_snp_tcb_version(get_tcb_version_from_vcek(x));
   memset(chipid, 0, CHIP_ID_SIZE);
   if (!get_chipid_from_vcek(x, chipid, CHIP_ID_SIZE)) {
-    printf(
-        "%s() error, line %d, Failed to retrieve HWID from VCEK extensions\n",
-        __func__,
-        __LINE__);
+    printf("%s() Info, line %d, Failed to retrieve HWID from VCEK extensions."
+           " Setting chip ID to 0.\n",
```
gapisback, line 430: Reworked message as suggested, and reduced severity to `Info`.

src/certifier.mak
```diff
@@ -1,8 +1,6 @@
 #    
 #    File: certifier.mak
 
-ENABLE_SEV=1
```
gapisback, line 4: Deleted this and moved this definition to start of `all:` target.

src/certifier.mak
```diff
@@ -92,7 +90,9 @@ CERTIFIER_LIB = certifier.a
 LIBCERTIFIER         = lib$(SWIG_CERT_INTERFACE)
 CERTIFIER_SHARED_LIB = $(LIBCERTIFIER).so
 
-all:	$(CL)/$(CERTIFIER_LIB)
+# Include SEV-SNP code-fragments, too, when building .a
+all: ENABLE_SEV=1
+all: $(CL)/$(CERTIFIER_LIB)
```
gapisback, line 95: With L94, we will still continue to compile-in code under `#if SEV_SNP` while building the static library.

We just will not pull in that code while building the shared library, which is being used just for pytests now.

Once the whole release model is better understood, we may have to revisit these Makefile-rules, but this arrangement seems acceptable to me (for now).

src/certifier.cc
gapisback: Discussed with @yelvmw : Get rid of message on L287; silently return `false`.

src/certifier.cc
gapisback: @yelvmw : Retain this message as Info; that's good enough.

src/certifier.mak
gapisback: With @yelvmw : remove this change, so the SEV_SNP code is included in shared libs, too.

gapisback: @yelvmw -- I have updated this change set as we discussed. Now SEV_SNP code chunks will be compiled with shared libraries, also.

I am ready to push this to `/main` once tests pass.

PR #188, opened 2023-08-26
Certifier in tee

jlmucb:
Most of certifier_in_tee code in place.  Still needs testing.

Squashed merge message:

Enable and demonstrate running a Certifier Service in a TEE

This adds support to simpleserver to: 
(a) use an encrypted policy key from a secure store, 
(b) act as a key server for the policy private key, 
(c) contact a key server to get its private policy key.  

There is a complete example in sample_apps/certifier_in_tee including
documentation.  There are also a few bug fixes and a note to help people 
who are planning real deployments after exploring the sample_apps.


gapisback:
@jlmucb -- Your CI job is failing due to some connectivity issue to `github.com` ... Possibly something is down somewhere.

I'm not able to run this script on my VM either from inside our VPN:

```
cc-sdb-vm:[53] $ pwd
/home/agurajada/Code/certifier-framework-for-confidential-computing/src/islet

cc-sdb-vm:[52] $ ../../third_party/islet/setup.sh
--2023-08-28 13:13:16--  https://github.com/Samsung/islet/archive/refs/tags/certifier-v1.0.1-beta.tar.gz
Resolving github.com (github.com)... 192.30.255.113
Connecting to github.com (github.com)|192.30.255.113|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://codeload.github.com/Samsung/islet/tar.gz/refs/tags/certifier-v1.0.1-beta [following]
--2023-08-28 13:13:16--  https://codeload.github.com/Samsung/islet/tar.gz/refs/tags/certifier-v1.0.1-beta
Resolving codeload.github.com (codeload.github.com)... 192.30.255.120
Connecting to codeload.github.com (codeload.github.com)|192.30.255.120|:443... connected.
HTTP request sent, awaiting response... 403 Forbidden
2023-08-28 13:13:16 ERROR 403: Forbidden.
```

certifier_service/certlib/certlib_policy.go
```diff
@@ -0,0 +1,440 @@
...
+	return true
+}
+
+func PolicyStoreDeleteEntry(ent int) {
+}
+
+func ConstructKeyForProtect(keyName string, keyType string) *certprotos.KeyMessage {
+
+	// only alg for now is "aes-256-cbc-hmac-sha256"
+	if keyType != "aes-256-cbc-hmac-sha256" {
```
gapisback, line 91: Here's another instance of a hard-code string for which there is an equivalent C-string constant in C-land.
I found more such instances when going thru remaining files.

We should log an issue to clean this up post-check-in ... if you agree. Otherwise, the hard-coded constants' creep will recur.

certifier_service/certlib/certlib_policy.go
```diff
@@ -0,0 +1,440 @@
...
+	k.KeyFormat = &format
+	k.SecretKeyBits = nakedKey
+	k.NotBefore = &nb
+	k.NotAfter = &na
+	return k
+}
+
+func ProtectBlob(enclaveType string, k *certprotos.KeyMessage, buffer []byte) []byte {
+
+	if k.GetKeyType() != "aes-256-cbc-hmac-sha256" {
```
gapisback, line 132: I found 3 instances of the hard-coded string `"aes-256-cbc-hmac-sha256"` in this file, alone.

I don't quite understand why we are only supporting just this one algorithm.  In any case, is it possible in Go to define a static string constant in this file? If yes, it will be better to create one named, say `Integrity_method_aes_256_cbc_hmac_sha256` (same as the one defined in `src/certifier_algorithms.cc`) and use that in all places in this Go-file.

And add some comments alongside the definition explaining why only this one algo is supported.

When we clean-up to use C-constant strings, the replacement should be a pass-thru.

Just a thought ... if this code consolidation is possible to be done for the new code added with this change.

gapisback (approved):
HI, @jlmucb -- I made a quick scan of this change-set, and scanned thru the Go-files

The bulk of this change is in Go Certifier Service and I didn't try to fully understand all the details. Am relying on @yelvmw to do a more close logic check review.

Some minor cleanup comments included ... some of which can be pushed off to post-check-in as follow-on issues.

certifier_service/certlib/cert1_test.go
```diff
@@ -286,17 +286,17 @@ func TestClaims(t *testing.T) {
...
-	   if  vat == nil {
-	           t.Errorf("attestation fails")
-	   }
-	   uvat :=  certprotos.Attestation{}
-	   err = proto.Unmarshal(vat, &uvat)
-	   if err != nil {
-	           t.Errorf("attestation unmarshal fails")
-	   }
-	   PrintAttestation(&uvat)
+		   fmt.Printf("\nAttest\n")
```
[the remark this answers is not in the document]
jlmucb: added comment

certifier_service/certlib/cert1_test.go
```diff
@@ -1826,3 +1852,177 @@ func TestGramineVerify(t *testing.T) {
...
+		t.Errorf("Stored key doesn't match")
+		return
+	}
+
+	fmt.Printf("\nRecovered key:\n")
+	PrintKey(newKey)
+}
+
+/*
+func TestTEESeal2(t *testing.T) {
```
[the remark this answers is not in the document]
jlmucb: This request requires initialization that is nopt currently done in the CI

src/support_tests.cc
```diff
@@ -105,7 +105,7 @@ bool test_encrypt(bool print_all) {
 bool test_authenticated_encrypt(bool print_all) {
   const int in_size = 2 * block_size;
   const int out_size = in_size + 256;
-  const int key_size = 64;
+  const int key_size = 96;
```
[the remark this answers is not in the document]
jlmucb: Because the sha-384 mac takes a longer key

sample_apps/certifier_in_tee/script
```diff
@@ -0,0 +1,120 @@
+#
+# Helpful script to run example
```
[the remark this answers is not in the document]
jlmucb: eventually

certifier_service/certlib/certlib_policy.go
jlmucb: fixed

certifier_service/certlib/certlib_policy.go
```diff
@@ -15,201 +15,73 @@
...
-    store_entry *             se = new store_entry();
-    entry_[i] = se;
-    se->tag_ = pe.tag();
-    se->type_ = pe.type();
-    se->value_ = pe.value();
-  }
-  num_ents_ = psm.entries_size();
-
-  return true;
+func PolicyStoreDeleteEntry(ent int) {
```
yelvmw, line 85: Looks like this is never implemented and used?

certifier_service/certlib/certlib_policy.go
```diff
@@ -290,15 +293,29 @@ func EncapsulateData(ek *certprotos.KeyMessage, alg string, data []byte, edm *ce
...
 	}
 
+	pK := rsa.PrivateKey{}
+	PK := rsa.PublicKey{}
+	if !GetRsaKeysFromInternal(ek, &pK, &PK) {
+		fmt.Printf("Can't recover keys\n")
+		return false
+	}
+
+	label := "huh"
```
yelvmw, line 303: What's this hard coded label? We have another one below.

yelvmw:
There are a lot changes in the series. I went through most of them but cannot sift through all the details in a short time. Other than some of the minor questions, I have the following two questions/observations:

1. We require a separate "key-provision" operation by invoking the certifier service before we can start it using the store. Does it not make sense to do the key provisioning once we failed to get the policy key from store and just use it if it succeeds?
2. We now added a new evidence purpose "key-provision". How do we prevent some app in the key policy domain from making the request and get the policy key? Maybe we decide this is not a concern. I can see that in theory for sure. But I want to bring it up so we at least can do some analysis of the trust relationship and deployment caveats.

certifier_service/certlib/certlib_policy.go
jlmucb: It is in simpleserver.go lines 805,838

certifier_service/certlib/certlib_policy.go
jlmucb: It's like a nonce, go crypto uses it to make sure the decrypt worked.

PR #196, opened 2023-09-18
Add support for NVIDIA H100 attestation

rgerganov:
This patch adds two interfaces for abstracting the NVIDIA platform and the NVIDIA GPU and some mock implementations.


src/test_data/attestationReport.txt
```diff
@@ -0,0 +1 @@
...0000000000003f013300013000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000400133000130000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001020ecb8f64dcfd50dabb760fc6a24210ba77d9f43a3a4db170a1f560283566f620106000800005e0096010000000e000400b10300000f00040038383200100002003000110005003130313000120005003032303000130002000000030007003534352e3030000400080080010000000000000c000001000000000100000004000000010000000100000001000000010000000100000000000000010000000000000082000000000000000000000000000000000000000000000000000000000000000000000000000000020000000200000000000000000000000100000001000000010000000000000001000000010000000100000001000000010000000100000001000000010000000100000001000000010000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000d00040000c001000b00010055f23c3889d1ecac1e082b0f59f8b498af309e22820b72dbcee7a18c2d8606fb462873c16c57698f0cf30f1b9360ff513c769f20507621678385382a19b6784f745c019441b6e96b0aaab5397d8826c27d9971f8e953b3e788b3582b1c96c38b80
```
yelvmw, line 1: Is this actually a verifiable report? Which key is used to sign it? If we have control, can we add nonce to the report generation?

yelvmw:
Looks good to me and thanks! I would put both the cc and header into src/nvidia or something. These should only be consumed internally by the certifier in the future anyway. It would look like src/sev-snp.

src/test_data/attestationReport.txt
rgerganov: Yes, this can be verified with the attestation certificate chain which is also added as test data. It is signed with the leaf key from the attestation chain. Unfortunately, we have to use a hardcoded nonce but I think it is still useful for testing purposes.

jlmucb (approved):

rgerganov:
I have finished the NVML implementation and this is ready for review.

jlmucb (approved):

Issue #194, opened 2023-09-08
SWIG/Python bindings issue with binary data containing Unicode chars.


gapisback:
Some of the key Certifier Framework interfaces cannot be invoked through SWIG-generated Python bindings due to the presence of surrogate Unicode characters in some data, such as certificates, policy-key etc.

A simple standalone repro has been extracted in this dev-environment:

- Branch: agurajada/repro-Py-Unicode-arg-bug
- Test: test_libswigpytests.py -k test_cc_trust_data_initialize_simulated_enclave_w_unicode_surrogate

----
Here is how the repro will fail:

```
PYTHONPATH=../.. pytest --capture=tee-sys -v test_libswigpytests.py -k test_cc_trust_data_initialize_simulated_enclave_w_unicode_surrogate

        # TypeError: in method 'cc_trust_data_initialize_simulated_enclave', argument 2 of type 'string const &'
>       result = cctd.initialize_simulated_enclave(attest_key_bin)

test_libswigpytests.py:106:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <swigpytests.cc_trust_data; proxy of <Swig Object of type 'swigpytests::cc_trust_data *' at 0x7fc8075f0c60> >
serialized_attest_key = b'\n\tattestKey\x12\x10rsa-2048-private\x1a\x07vse-key"\x9a\t\n\x80\x02\xb9\x928\xa1\xd8\x9c\xb2\x12\'\xb97\xa5\xab\xd...\xf6\xb5\xaak\x86\x06\x1a\x1f{\xa74\xd2\xa1\x0f\x84\xd0^\xbb\x8a\xd6\xadF\xef\x93\xb1\x17\xa9\xa5\x07\xac\xc5\xb0H\xbe'

    def initialize_simulated_enclave(self, serialized_attest_key):
>       return libswigpytests.cc_trust_data_initialize_simulated_enclave(self, serialized_attest_key)
E       TypeError: in method 'cc_trust_data_initialize_simulated_enclave', argument 2 of type 'string const &'

../../swigpytests.py:75: TypeError
==================================================== short test summary info ====================================================
[...]

PR #197, opened 2023-09-22
(#194) Fix Python binding interfaces to work with Unicode chars

gapisback:
As part of previous API changes (@SHA 1d2ff1a), a collection
of Python tests were disabled as the bindings and calling
interfaces were running into errors in the presence of
Unicode surrogate characters. These tend to appear in
certificates, keys and other binary data.

This commit resolves these interfaces and calling conventions
to re-enable failing test cases. Summary of the nature of fixes
applied:

- Add Python-specific interfaces for few methods which will be
  used by existing pytests. The plan is to add similar
  python_<...> interfaces as needed for developing simple-app.

- Define SWIG-interface rules to manage Python bindings:
  - Use pybuffer_binary() to invoke interfaces using Python
    byte-streams (rather than string arguments)
  - Add 'ignore' directives, for over-loaded C++ interfaces

- Implement a post-SWIG-generation shell script to do in-place
  replacement of generated code. (This works-around the item
  listed in SWIG issue #1916.) This script is run as part of
  the makefile build rules that invoke SWIG.

--------

**NOTE: To reviewers**: I will update the commit message later. Basically, this change-set has a small collection of fixes and tweaks to overcome the passing of parameters containing Unicode data with surrogate characters via Python interfaces. All but one of the existing pytests that were commented out as part of 1d2ff1a (API-changes) have now been re-enabled and should be executing cleanly.

When possible, existing interfaces that took string arguments have been overloaded with a Python-specific interface taking `(byte *, int)` pair of arguments.

Some minor SWIG interfaces changes are done to help reduce disambiguation confusion while invoking interfaces.

One test, test_certifier_framework.py::test_run_app_as_a_server() consistently fails in CI whereas it passes cleanly on my Nimbus-VM env. For now, this test case stays skipped.


.github/workflows/build.yml
```diff
@@ -124,21 +124,20 @@ jobs:
...
 
         # Just run the basic case of this test to see classes created in generated Python module
-         PYTHONUNBUFFERED=TRUE PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python \
+        PYTHONUNBUFFERED=TRUE PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python \
              pytest --capture=tee-sys -v test_certifier_protobuf_interfaces.py \
                  -k test_certifier_pb2_basic
 
         # Re-run entire test w/o capturing output
-         PYTHONUNBUFFERED=TRUE PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python \
+        PYTHONUNBUFFERED=TRUE PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python \
```
gapisback, line 136: All of the above are editing fixes to fix alignment.

.github/workflows/build.yml
```diff
@@ -124,21 +124,20 @@ jobs:
...
              pytest -v test_certifier_protobuf_interfaces.py
 
         # Exercise SWIG-Python bindings standalone unit-tests
         PYTHONUNBUFFERED=TRUE pytest --capture=tee-sys -v test_libswigpytests.py
-        PYTHONUNBUFFERED=TRUE pytest -v test_certifier_framework.py
```
gapisback, line 141: Deleted because this was added in previous commit, when several existing pytests were commented out as they were failing.

The pytests are broken up into two sets: Those that run w/o Cert Service. And those that need Cert Service to be up. This L141 will attempt to run the latter kind, too, when Cert Service is still not up. And they will fail.

Delete this line as the latter set of tests are being run later on, downstream.

include/certifier_framework.h
```diff
@@ -234,11 +234,20 @@ class cc_trust_manager {
 
   // Each of the enclave types have bespoke initialization
 
+  // Interface ignored through SWIG bindings
```
gapisback, line 237: The `%ignore` directive comes in `certifier_framework.h` .

include/certifier_framework.h
```diff
@@ -362,16 +371,16 @@ class secure_authenticated_channel {
                        key_message & private_key,
                        const string &private_key_cert);
 
+  bool init_client_ssl(const string &          host_name,
```
gapisback, line 374: No change; just code movement to keep similar interfaces nearby.

include/certifier_framework.i
```diff
@@ -13,9 +13,19 @@
...
 
-// Needed to invoke init_policy_key() using Python bytestream arg.
+// Needed to invoke init_policy_key() and other interfaces that take
+// (byte *, int) supplied by Python bytestream arg.
 %include <pybuffer.i>
-%pybuffer_binary(byte* asn1_cert, int asn1_cert_size)   // cc_trust_data()->init_policy_key()
+
+%pybuffer_binary(const byte *serialized_attest_key, int attest_key_size);
+%pybuffer_binary(const byte *measurement, int measurement_size);
+%pybuffer_binary(const byte *serialized_attest_endorsement, int attest_key_signed_claim_size);
```
gapisback, line 22: These 3 directives are to marshall the parameters of `initialize_simulated_enclave()` as byte-stream arguments.

include/certifier_framework.i
```diff
@@ -13,9 +13,19 @@
...
+
+%pybuffer_binary(byte* asn1_cert, int asn1_cert_size);   // cc_trust_data()->init_policy_key()
+
+%ignore initialize_simulated_enclave(const string &serialized_attest_key,
```
gapisback, line 26: This is being ignored; otherwise, we have over-loaded interfaces. 

In the SWIG-generated wrap.cc file, the logic to pick the right interface is not working as expected. The code-flow gets confused and we get an error that the arguments are incorrect.

By ignoring this interface, we leave just the interface with triplets of byte-stream interfaces as the valid Python interface to invoke. This allows tests to work cleanly.

src/certifier.mak
```diff
@@ -84,11 +83,13 @@ SWIG_CERT_INTERFACE = certifier_framework
...
 SWIG_PYTEST_INTERFACE = swigpytests
 
+FIX_SWIG_SCRIPT = $(CERTIFIER_ROOT)/CI/scripts/fix_swig_wrap.sh
+
 PY_INCLUDE = -I /usr/include/python3.10/
 
 #export LD_LIBRARY_PATH=/usr/local/lib
 LDFLAGS = -L $(LOCAL_LIB) -lprotobuf -lgtest -lgflags -lpthread -L/usr/local/opt/openssl@1.1/lib/ -lcrypto -lssl
-LDFLAGS_SWIGPYTEST = -L $(LOCAL_LIB)
+LDFLAGS_SWIGPYTEST = -L $(LOCAL_LIB) -l protobuf
```
gapisback, line 92: Some SWIG-unit-tests now use protobuf classes in the interfaces, hence need to link with protobuf libraries while building `swigpytestssharedlib` library.

src/swigpytests.cc
```diff
@@ -31,6 +32,47 @@ swigpytests::cc_trust_data::cc_trust_data() {
 
 swigpytests::cc_trust_data::~cc_trust_data() {}
 
+bool swigpytests::cc_trust_data::initialize_simulated_enclave(
```
gapisback, line 35: Method copy-cloned and added to SWIG-unit-test exerciser, to verify that the SWIG mapping [etc.] is working right.

tests/pytests/test_certifier_framework.py
```diff
@@ -497,33 +503,39 @@ def cc_trust_manager_get_certified(cctd):
...
     with open(example_app_measurement_file_name, 'rb') as example_app_measurement_file:
         example_app_measurement = example_app_measurement_file.read()
 
+    attest_endorsement_file_name      = CertPyTestsDir + '/data/platform_attest_endorsement.bin'
     with open(attest_endorsement_file_name, 'rb') as attest_endorsement_file:
         platform_attest_endorsement_bin = attest_endorsement_file.read()
 
-    result = cctd.initialize_simulated_enclave(str(attest_key_bin),
-                                               str(example_app_measurement),
-                                               str(platform_attest_endorsement_bin))
```
gapisback, line 518: We don't need this manual "string-ification" here ... Instead, see L519.

tests/pytests/test_certifier_framework.py
```diff
@@ -497,33 +503,39 @@ def cc_trust_manager_get_certified(cctd):
...
+    result = cctd.initialize_simulated_enclave(attest_key_bin,
+                                               example_app_measurement,
+                                               platform_attest_endorsement_bin)
```
gapisback, line 530: This call invokes the interface using byte-streams that are established above using `read()` with `open(... rb)`.

jlmucb (approved):

include/certifier_framework.h
gapisback: Note: Cross-check if with the ignore directive do we still need this shell script.

include/certifier_framework.h
```diff
@@ -234,11 +234,20 @@ class cc_trust_manager {
...
   // Each of the enclave types have bespoke initialization
 
+  // Interface ignored through SWIG bindings
   bool initialize_simulated_enclave(
       const string &serialized_attest_key,
       const string &measurement,
       const string &serialized_attest_endorsement);
 
+  // Interface invoked through Python apps
+  bool initialize_simulated_enclave(const byte *serialized_attest_key,
```
gapisback, line 244: @jlmucb said: If you can remove the overloading by renaming this as `python_initialize...`, that is perfectly fine too.

We can then get rid of the hacky shell script Perl replacement logic.

CI/scripts/fix_swig_wrap.sh
```diff
@@ -0,0 +1,37 @@
...
+# the workaound-fix in user-defined typemap()-ing code ran into other
+# compilation issues in the generated code.
+# This replacement is needed for these reasons:
+#
+#  - Even though the C++ variants of certain overloaded interfaces like
+#    init_client_ssl(), init_server_ssl() are ignored, the data for
+#    certain arguments [e.g., string &asn1_root_cert in these methods]
+#    contains Unicode surrogate chars. We need to apply this workaround-fix
+#    in order for the Python invocation to even succeed argument passing,
+#    even with the %ignore specifier in the SWIG.i interface file.
```
gapisback, line 26: @yelvmw -- Note: This point is new since the last time we visited this packet on Monday.

I have discovered thru unit-testing that for Python bindings of both the public `certifer_framework.h` and unit-testing `swigpytests.h`, we do need this shell script. 

Just using SWIG's `%ignore` directive alone to avoid overloaded interfaces is not sufficient to get all scenarios to work correctly.

yelvmw: Would like to know a bit more of the details here. But if it's necessary, I won't block the pull request.

include/certifier_framework.i
```diff
@@ -13,9 +13,34 @@
...
+// ----------------------------------------------------------------------------
+%ignore initialize_simulated_enclave(const string &serialized_attest_key,
+                                     const string &measurement,
+                                     const string &serialized_attest_endorsement);
+
+%ignore init_client_ssl(const string &host_name, int port,
+                        const cc_trust_manager &mgr);
+
+%ignore init_server_ssl(const string &host_name, int port,
+                        const cc_trust_manager &mgr);
```
yelvmw, line 43: You sure we could ignore these for the Python simple app?

yelvmw (approved):
Some things went over my head. Mostly the SWIG complexities. TBH, I don't like the renaming for Python. Given that there is no better solution, I guess we'll have to live with it...

CI/scripts/fix_swig_wrap.sh
gapisback: @yelvmw -- I can talk to you offline to help clarify these gyrations.

Thru unit-testing, I found that we need both the `%ignore` directive in the SWIG-interface file -and- also the patch bash script to do the perl string replacement.

include/certifier_framework.i
gapisback: Hmmm ... Thanks for spotting this @yelvmw -- This made me do some more thinking and debugging / testing investigation.

I made the following changes in response to your comment:

- These two `ignore` directives are not needed for tests in `test_certifier_framework.py` to run cleanly.
- So commented them out, but left them behind. Added an elaborate comment explaining when / why we might need these to be re-enabled. (For future reference; I'm going to forget all this stuff ...)
- Added couple of new test cases to verify that these interfaces can still be invoked with `const cc_trust_manager &mgr` argument . And that the tests will still succeed. 
   - Add test case `test_run_app_as_a_client_init_client_ssl_with_trust_manager()`
   - Add test case `test_run_app_as_a_server_with_trust_manager()`
 
- Figured the an alternate fix to avoid the 120 seconds sleep in some test cases. In the method `server_dispatch()`, for the testing-hook when call-back function hdlr is NULL, also do `close((sock);` before exiting. This should take away the unpredictability seen in CI runs. (Will re-test and verify.)

PR #198, opened 2023-09-26
Refactor protobuf messages

rgerganov:
Introduce new protobuf messages which encapsulate various evidence packages which are being sent to the Certifier Service. Using separate message types for each platform is less error-prone and allows extensibility.


certifier_service/certprotos/certifier.proto
```diff
@@ -247,15 +247,53 @@ message proof {
...
+  optional bytes attestation                = 2;
+};
+
+enum msg_purpose {
+  authentication                            = 0;
+  attestation                               = 1;
+};
+
 message trust_request_message {
   optional string requesting_enclave_tag    = 1;
```
rgerganov, line 286: why do we need this and the next one?

jlmucb: yes

jlmucb (changes requested):
I would not refactor it this way right now.  The evidence packages can change, for example, there could be a cert chain for some of the platforms.  Also, there not be one evidence package format even for a single platform.  (i.e.- someone could specify rules in an alternative format even for the same platform).

I like the idea but could we talk before doing this?

Are these changes needed for some functional reason or just to beautify?

rgerganov:
This refactoring will enable adding support for NVIDIA H100. The current way of sending evidence to the Certifier Service is very error-prone and I don't see why we should keep doing it like that:
 * using hard-coded strings for evidence type is bad; these strings are not part of the protocol definition and need to be hard-coded in both C++ and Go
 * relying on implicit list order is even worse; for SEV-SNP there is an implicit assumption that certificates are ordered ARK,ASK,VCEK in the evidence package and there is no way find this out from the protocol definition
 * using string literals instead of enums is also bad practice

My proposal is to fix this before adding more technical debt by following the current pattern for new platforms like NVIDIA.

PR #200, opened 2023-10-02
Fix Python include dirs

rgerganov:
Do not hardcode Python include dirs, use pkg-config to set them appropriately.


gapisback (approved):
Seems like a simple one-liner fix. Thanks for proactively doing this fix.

gapisback:
HI, @rgerganov -- I have passed this one-line change fix. Did this get lost in the shuffle? Can you please check this in so we get the cleanup done soon. Thanks!

Issue #156, opened 2023-07-19
Develop simple_app in Python using swig-generated Python bindings.


gapisback:
Recent check-in PR #144 adds support for swig-generated Python bindings. Using this infrastructure, develop simple_app written in Python to replicate exactly the same functionality as is provided under `sample_apps/simple_app/example_app.cc`.

gapisback:
Bulk of simple-app in Python has been developed and merged under PR #202

PR #202, opened 2023-10-04
(#156) Add simple_app in Python, using SWIG-generated Python bindings

gapisback:
This commit adds a nearly-complete simple_app written in Python
using SWIG-generated Python bindings. Basic interfaces such as
init_policy_key(), initialize_simulated_enclave(), cold_init(),
warm_restart() and certify_me() are functional.

The missing pieces are to establish secure channel connection
between client / server portions of "the app" and to communicate
secure messages. That will be applied in future commits.


sample_apps/run_example.sh
```diff
@@ -135,6 +135,7 @@ SampleApps=( "simple_app"
              "simple_app_under_app_service"
              "simple_app_under_keystone"
              "simple_app_under_islet"
+             "simple_app_python"
```
gapisback, line 138: Most of the changes in this script is to inject this new app in existing switch / case code-flow.

sample_apps/run_example.sh
```diff
@@ -287,7 +288,7 @@ Steps=( "rm_non_git_files"
         # "do_cleanup"
         "build_utilities"
         "gen_policy_and_self_signed_cert"
-        "embed_policy_in_example_app"
+        "gen_policy_key_for_example_app"
```
gapisback, line 291: Renamed func as it's more accurate.

sample_apps/run_example.sh
```diff
@@ -809,13 +812,27 @@ function setup_cca_emulated_islet_shim_bins() {
...
     run_pushd "${PROV_DIR}"
 
+    python_flag=""
+    init_cert_flag=""
+    outfile="policy_key.cc"
+
+    # Invoke utility with python-specific arguments.
+    # Use upper-case names to avoid pylint errors.
+    if [ "${SampleAppName}" = "simple_app_python" ]; then
+        python_flag="--python"
```
gapisback, line 826: The `embed_policy_key.exe` has been enhanced to take `--python`-specific flags which changes the format in which output is rendered.

sample_apps/simple_app_python/example_app.py
```diff
@@ -0,0 +1,318 @@
+#!/usr/bin/env python3
+################################################################################
+# main.py
+################################################################################
+"""
+Python simple-app to demonstrate use of Certifier Framework APIs.
```
gapisback, line 6: This is the new Python sample-app.

tests/measurement_utility_test.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
+# #############################################################################
+# Simple shell script to 'test' that the measurement generated by the
```
gapisback, line 3: New test script to validate that the `--policy-key` argument to `embed_policy_key.exe` is working correctly.

During dev, I had made an error in the program, not including the 2nd file's contents. This went undiscovered till something else broke in the simple_app dev steps. Hence, I added this little test-case to protect us from such silly mistakes in the future.

jlmucb (approved):

tests/pytests/data/example_app.measurement
```diff
@@ -1 +1 @@
-�������ݒ��c�u����uԝ��47�
```
rgerganov, line 1: if this file is generated during tests, you can add it to .gitignore

rgerganov (commented):
Look good, some minor comments inline.

yelvmw (approved):
LGTM. Not for this change, but we might want to make SEV the default for Python and provide a JSON policy. I think this is most likely what people would be looking for in practice.

.github/workflows/build.yml
```diff
@@ -214,6 +214,17 @@ jobs:
...
 
+        echo " "
+        echo "---- Running measurement_utility_test.sh ... ----"
+        echo " "
+        # Basic verification of measurement_utility.
+        #
+        ./tests/measurement_utility_test.sh
+
+        # Above script will execute a utility which will generate policy_key.py
+        git restore sample_apps/simple_app_python/policy_key.py
```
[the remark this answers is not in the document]
gapisback: I kicked myself for not thinking about this approach, and so tried to implement this.

However, when implemented, this does not work. `.gitignore` will only ignore files that are not already committed to git. See this:

```
A gitignore file specifies intentionally untracked files that Git should ignore. Files already
       tracked by Git are not affected; see the NOTES below for details.
```

I recall, now, why I needed this file checked-in to git.

In this case, I do need this file in Git, even as just a dummy file. 

Otherwise, if this file is not in git, in a new branch, any attempts to run even just `./example_app.py --help` will fail as follows:

```
cc-sdb-vm:[18] $ PYTHONPATH=../.. ./example_app.py --help
Traceback (most recent call last):
  File "/home/agurajada/Code/certifier-framework-for-confidential-computing/sample_apps/simple_app_python/./example_app.py", line 13, in <module>
    import policy_key
ModuleNotFoundError: No module named 'policy_key'
```

I may be able to overcome this with some Python-gyrations, to avoid failing even when this file `policy_key.py` is missing. But I didn't think that was all that much more a better solution.

Hence, this file remains checked-in and I cannot `.gitignore` it.

tests/pytests/data/example_app.measurement
gapisback: OK, this one made sense and seemed to work when implemented. I took the opportunity to clean-up more generated files by adding them to `.gitignore`.

Will re-test to make sure things still run in CI ...

sample_apps/simple_app_python/example_app.py
```diff
@@ -0,0 +1,318 @@
...
+                     'for purpose of', purpose, '\n')
+        sys.exit(1)
+
+    public_key_alg = "rsa-2048"
+    symmetric_key_alg = "aes-256-cbc-hmac-sha256"
+
+    # -------------------------------------------------------------------------
+    # Should succeed with valid key algorithm names, after policy key has been
+    # initialized
+    if operation == 'cold-init':
```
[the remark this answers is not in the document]
gapisback: @rgerganov -- Even while developing this program, I did consider modularizing inline code into sub-functions.

It seemed like needless refactoring. The body of each block is currently fairly small. To refactor them into functions, I'd have to deal with passing-down arguments and so on. Can be done; but just seemed a bit risky at this late stage. The way it's coded now, one benefit is that the scope of Python objects and CC stuff is very obvious as everything is in-line and at the same "code-block" level.

If you don't insist, I'll skip this rework, so that I can get this in (tomorrow) after you've given this PR a final look.

tests/pytests/data/example_app.measurement
gapisback: @rgerganov -- I had to re-instate the existing files in `tests/pytests/data` dir. I could not remove them as some existing pytests are written to work off of these checked-in files. (CI job failed miserably otherwise.)

So the only change arising from your comment is some cleanup in `.gitignore`. As that was simple enough, I squashed all post-review commit into the main one. You can just take a re-look at this one .gitignore file.

sample_apps/simple_app_python/example_app.py
```diff
@@ -0,0 +1,318 @@
...
+    pf_file_name            = parsed_args.pf_file_name
+    policy_host             = parsed_args.policy_host
+    policy_port             = parsed_args.policy_port
+    policy_store_file       = parsed_args.policy_store_file
+    server_app_host         = parsed_args.server_app_host
+    server_app_port         = parsed_args.server_app_port
+    print_all               = parsed_args.print_all
+
+    if print_all:
+        print('INITIALIZED_CERT_SIZE = ', policy_key.INITIALIZED_CERT_SIZE)
```
gapisback, line 70: @rgerganov -- From your comment in the draft PR that this variable is not needed -- I agree.

In code, I do use `len()` when needed. 

I felt it's just a small enough change that keeps the output in the .py and .cc file be consistent. 

Hence, I retained the generation of this `INITIALIZED_CERT_SIZE` variable.

.github/workflows/build.yml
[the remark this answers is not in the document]
rgerganov: One simple solution is to put the import statement into `try..except` block and print an error message saying that `policy_key.py` needs to be generated. That way we won't need to put generated files into git.

This can be addressed later though, so go ahead and merge this PR. Thanks.

sample_apps/simple_app_python/example_app.py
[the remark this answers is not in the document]
rgerganov: Fair enough.

.github/workflows/build.yml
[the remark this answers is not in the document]
gapisback: @rgerganov -- I know you said we can try to do this in a follow-on, but I think it's going to be very tedious for the user to "figure out" how to generate `policy_key.py`.

Looking at `--dry-run` output of `run_example.sh`, the message will have to instruct the user to do two things:

- First, run `utilities/cert_utility.exe` to generate `--operation=generate-policy-key-and-test-keys ...  --policy_cert_output_file=policy_cert_file.bin`
- Then, run `utilities/embed_policy_key.exe` to consume as `--input=policy_cert_file.bin`, the generated policy-cert-file
- This will produce the required `--output=../policy_key.py`

I think this is very user un-friendly to have to go thru multiple gyrations so that they can use simple stuff like `--help` interface.

I think we probably just have to live with this dummy file checked-in to git.

PR #203, opened 2023-10-04
Accelerator verification

jlmucb:


certifier_service/certprotos/certifier.proto
```diff
@@ -247,15 +253,20 @@ message proof {
   repeated proof_step steps                 = 3;
 };
 
+message accelerator_group {
+  optional string type                      = 1;
```
rgerganov, line 257: we may need to include other fields like uuid, driver version, vbios version, etc.

include/certifier_framework.h
```diff
@@ -144,6 +144,23 @@ class domain_info {
   int    service_port_;
 };
 
+const int max_accerlerators = 4;
+class accelerator {
+public:
+  accelerator();
+  ~accelerator();
+  string accelerator_type_;
```
rgerganov, line 152: this should be enum

include/certifier_framework.h
```diff
@@ -144,6 +144,23 @@ class domain_info {
...
+  bool verified_;
+  string location_type_;  // in-memory, network, file
```
rgerganov, line 154: this should be enum

include/certifier_framework.h
```diff
@@ -144,6 +144,23 @@ class domain_info {
...
+  string file_name;
+  string network_address_;
+  byte* address_;
+  long  size_;
+  int num_certs_;
+  string* certs_;
```
rgerganov, line 160: use `std::vector`

include/certifier_framework.h
```diff
@@ -144,6 +144,23 @@ class domain_info {
...
 
+const int max_accerlerators = 4;
+class accelerator {
+public:
+  accelerator();
+  ~accelerator();
+  string accelerator_type_;
+  bool verified_;
+  string location_type_;  // in-memory, network, file
+  string file_name;
```
rgerganov, line 155: what kind of `file_name` or `network_address` is associated to GPU?

include/certifier_framework.h
```diff
@@ -163,6 +180,11 @@ class cc_trust_manager {
   string public_key_algorithm_;
   string symmetric_key_algorithm_;
 
+  int num_accelerators_;
+  accelerator  accelerators_[max_accerlerators];
```
rgerganov, line 184: use `std::vector`

src/cc_helpers.cc
```diff
@@ -338,6 +353,18 @@ bool certifier::framework::cc_trust_manager::initialize_oe_enclave(
 #endif
 }
 
+bool certifier::framework::cc_trust_manager::accelerator_verified(const string& acc_type) {
+  for (int i = 0; i < num_accelerators_; i++) {
+    if (acc_type == accelerators_[i].accelerator_type_)
```
rgerganov, line 358: what if we have several accelerators of the same type? for example several NVIDIA H100

jlmucb:
Protobuf and code changes to support accelerators like GPUs when they are implemented

Issue #214, opened 2023-10-25
Rework handling of private-key in simple_app_python


gapisback:
As part of the commit @ SHA 68b0376, `simple_app_python/example_app.py` has been enhanced to implement secure SSL channel communication between client and server "apps", which use certificate and private-keys generated by the Certifier Framework.

In order to conform to Python's SSL interfaces, these certificate & keys need to be stored in files which the interfaces uses as below (from above sample program):

```
237 ###############################################################################
238 def server_dispatch(data_dir, server_app_host, server_app_port):
239     """
240     Start a server process listening on a SSL socket waiting for client input.
241     """
242     print(fnl(), 'Start server process ...\n')
[...]
249     key_outfilename  = os.path.join(data_dir, APPLN_KEYF_SIGNED_BY_ROOT_CA)
250     context.load_cert_chain(certfile = cert_outfilename, keyfile = key_outfilename)
```

The client-side code looks similar.

In order to call `context.load_cert_chain()`, on L250, the private-key for the certificate is stashed away on a file in the filesystem, here:

```
337 ###############################################################################
338 def dump_private_key_to_file(cctm, outfile, print_all):
[...]
349     result = cctm.write_private_key_to_file(outfile)
```

Writing out the private-key to a disk-resident file is a short-term implementation just to get an end-to-end working demo of this simple_app using Python SSL interfaces.

Under this issue, figure out a secure way to rework the handling of this private key.

A related issue to be covered in this issue is the following: Currently, the certificate & Key are persisted to disk as part of the `get-certified` target. The scenario to also cover is this one:

```
173     elif operation == 'run-app-as-server':
174         print(operation)
175
176         result = cctm.warm_restart()
177         if result is False:
178             print(fnl(), 'warm_restart() failed\n')
179             sys.exit(1)
180
181         result = server_dispatch(PROVISIONING_DIR,
182                                  server_app_host, server_app_port)
```

After `cctm.warm_restart()` the application should still be able to access the required certificate / private-key so that the SSL-channel is still established correctly.

gapisback:
HI, @jlmucb @yelvmw @rgerganov - I have done some [ more ] investigation on the Python SSL support to invoke interfaces using certificates & keys from in-memory data, rather than from files.

[...]

PR #221, opened 2023-11-03
(#214) Persist private-key data to named temp files. Rework warm_restart() execution.

gapisback:
This commit re-works the way private-key data is managed to be able to call Python SSL interfaces that require certificate and private-key data to be read from files. 

In Python `example_app.py`, private-key data that is instantiated in the `cc_trust_manager{}` object by reading it from the policy-store, will be written to named temp files, in anonymous but-named temp dirs. 

The private-key data will exist **very briefly in the tmp-files** while the SSL certificate-related method is invoked to load-certs.

With this change, `example_app.py` also now handles the case when trust-data is re-loaded from the policy-store by a warm-restart operation. 

Existing pytests have been refactored and extended to add a new pair of client/server tests going thru secure-SSL communication exchange, working off private-key data that is [now] persisted to tmp-files.

`test.sh` is extended to invoke this new pair of secure-channel test-cases. Existing `simple_app_python` test-execution  target is extended to exercise a variation where client-and-server can be restarted, without the Certifier Service, to verify the re-establishment of secure data thru `warm_restart()`.

----
**NOTE**: 

See the internal engineering analysis doc, here: [2023-11-CF-In-Memory-Private-Key-Support](https://onevmw.sharepoint.com/:w:/r/teams/VSE/Shared%20Documents/General/Internal/Engineering/2023-11-CF-In-Memory-Private-Key-Support.docx?d=w90ea145aed7e47938352896953be4fa7&csf=1&web=1&e=l6WFY6)
  where the gory details of Python SSL support efforts done in the public domain have been traced & documented.
  
  This doc gives the reasons why the only solution available to us now is the use of named temporary files.

As of this writing (Nov 2023), there is no support in Python OpenSSL interfaces to load certificates or private-keys
directly from in-memory data. We have to invoke SSL interfaces using files containing such data. Based on
the feedback on one of the threads listed in the above doc, the solution implemented for issue #214 is the use of very short-lived named temp-files created in anonymous but-named temp-dirs.

### DIFF
--- a/CI/scripts/test.sh
+++ b/CI/scripts/test.sh
@@ -202,6 +202,7 @@ function test-mtls-ssl-client-server-comm-pytest() {
 
     cd pytests
 
+    # ----
     # Exercise client-server communication with self-signed certificates
     pytest --capture=tee-sys -v test_client_server_mtls.py -k test_server_process_with_mtls_self_signed_cert &
 
@@ -209,15 +210,31 @@ function test-mtls-ssl-client-server-comm-pytest() {
 
     pytest --capture=tee-sys -v test_client_server_mtls.py -k test_client_app_with_mtls_self_signed_cert
 
+    # ----
     # Basic test to exercise core certificate verification v/s root-CA-cert
     pytest --capture=tee-sys -v test_client_server_mtls.py -k test_verify_certs_versus_root_cert
 
+    # ----
     # Exercise client-server communication with root-CA-signed certificates
-    pytest --capture=tee-sys -v test_client_server_mtls.py -k test_server_process_with_mtls_root_signed_cert &
+    pytest --capture=tee-sys -v test_client_server_mtls.py \
+            -k test_server_process_with_mtls_root_signed_cert_and_pvt_key_file &
 
     sleep 5
 
-    pytest --capture=tee-sys -v test_client_server_mtls.py -k test_client_app_with_mtls_root_signed_cert
+    pytest --capture=tee-sys -v test_client_server_mtls.py \
+            -k test_client_app_with_mtls_root_signed_cert_and_pvt_key_file
+
+    # ----
+    # Exercise client-server communication with root-CA-signed certificates
+    # using private-key written-to and read-from a temp-file, for certificate
+    # verification.
+    pytest --capture=tee-sys -v test_client_server_mtls.py \
+            -k test_server_process_with_mtls_root_signed_cert_using_temp_pvt_key_file &
+
+    sleep 5
+
+    pytest --capture=tee-sys -v test_client_server_mtls.py \
+            -k test_client_app_with_mtls_root_signed_cert_using_temp_pvt_key_file
 
     popd > /dev/null 2>&1
 }
@@ -480,7 +497,35 @@ function test-run_example-simple_app_python() {
     ./run_example.sh simple_app_python setup
     ./run_example.sh simple_app_python run_test
 
+    # ---- Variation -----------------------------------------------------------
+    # Test the scenario that once certified, the server-process and client-app
+    # can be re-started multiple times, without need for the Certifier Service.
+    # Re-starting the server-process will re-load the trust data from the
+    # policy-store. It will no longer need to contact the Certifier Service
+    # to get re-certified.
+    # Then, you can re-start the client-app, to talk to this now-certified
+    # server-process directly through secure SSL channel.
+    # The following sequence of steps validate this workflow and verify that
+    # the client and server can communicate through this secure channel.
+    # --------------------------------------------------------------------------
+    echo " "
+    echo "*******************************************************************"
+    echo "* Test Variation: Re-start server-process and client-app after shutting down the Certifier Service ..."
+    echo "*******************************************************************"
+    echo " "
+
     ./cleanup.sh
+
+    # This invokes the 'run-app-as-server' operation in example_app.py
+    # which does a warm_restart() to re-load trust data from the policy-store.
+    ./run_example.sh simple_app_python run_app_as_server_offers_trusted_service
+
+    # This invokes the 'run-app-as-client' operation in example_app.py
+    # which does a warm_restart() to re-load trust data from the policy-store.
+    # There is assert code way-deep-down in example_app.py that verifies that
+    # the client gets an expected 'Hello' message from the server, through this
+    # exchange.
+    ./run_example.sh simple_app_python run_app_as_client_make_trusted_request
 
     popd > /dev/null 2>&1
 }
--- a/sample_apps/simple_app_python/example_app.py
+++ b/sample_apps/simple_app_python/example_app.py
@@ -299,9 +299,8 @@ def setup_server_ssl_context(cctm, data_dir:str, print_all:bool):
     cert_outfilename = os.path.join(data_dir, APPLN_CERT_SIGNED_BY_ROOT_CA)
 
     with TemporaryDirectory() as tempdir:
-        with NamedTemporaryFile('wt', dir=tempdir, delete=False) as temp_keyf:
-            temp_keyf.close()
-
+        # With Python v3.10, can also add: delete_on_close=True (default)
+        with NamedTemporaryFile('wt', dir=tempdir, delete=True) as temp_keyf:
             dump_private_key_to_file(cctm, temp_keyf.name, 'server', print_all)
 
             context.load_cert_chain(certfile = cert_outfilename,
@@ -385,9 +384,8 @@ def setup_client_ssl_context(cctm, data_dir:str, print_all:bool):
     cert_outfilename = os.path.join(data_dir, APPLN_CERT_SIGNED_BY_ROOT_CA)
 
     with TemporaryDirectory() as tempdir:
-        with NamedTemporaryFile('wt', dir=tempdir, delete=False) as temp_keyf:
-            temp_keyf.close()
-
+        # With Python v3.10, can also add: delete_on_close=True (default)
+        with NamedTemporaryFile('wt', dir=tempdir, delete=True) as temp_keyf:
             dump_private_key_to_file(cctm, temp_keyf.name, 'client', print_all)
 
             context.load_cert_chain(certfile = cert_outfilename,
--- a/sample_apps/simple_app_python/example_app.py
+++ b/sample_apps/simple_app_python/example_app.py
@@ -13,8 +13,8 @@ import argparse
 from tempfile import TemporaryDirectory, NamedTemporaryFile
 from inspect import currentframe
 
-import policy_key
 import certifier_framework as cfm
+import policy_key
 
 ###############################################################################
 # Global Variables: Used in multiple places. List here for documentation
@@ -26,7 +26,6 @@ THIS_SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
 # Script defaults
 ATTEST_KEY_FILE                 = 'attest_key_file.bin'
 APP_DATA_DIR                    = './app1_data/'
-PROVISIONING_DIR                = './provisioning'
 EXAMPLE_MEASUREMENT             = 'example_app.measurement'
 APP_OP_TYPES                    = 'cold-init, get-certified, run-app-as-client, run-app-as-server'
 PLATFORM_ATTEST_ENDORSEMENT     = 'platform_attest_endorsement.bin'
@@ -36,12 +35,6 @@ POLICY_HOST                     = 'localhost'
 POLICY_HOST_PORT                = 8123
 SERVER_APP_HOST                 = '127.0.0.1'
 SERVER_APP_PORT                 = 8124
-
-# ------------------------------------------------------------------------------
-# Script global symbols to app-data dirs.
-CLIENT_APP_DATA='app1_data'
-SERVER_APP_DATA='app2_data'
-PROVISIONING_DIR='./provisioning'
 
 # Basenames for output files created to persist certificates and keys
 APPLN_CERT_SIGNED_BY_ROOT_CA = 'appln.PEM.cert'
@@ -173,8 +166,8 @@ def do_main(args) -> bool:
             print(fnl(), 'warm_restart() failed\n')
             sys.exit(1)
 
-        write_certficates_to_file(cctm, 'server', PROVISIONING_DIR, print_all)
-        result = server_dispatch(cctm, PROVISIONING_DIR,
+        write_certificates_to_file(cctm, 'server', appln_data_dir, print_all)
+        result = server_dispatch(cctm, appln_data_dir,
                                  server_app_host, server_app_port, print_all)
         if result is False:
             print(fnl(), 'server_dispatch() failed\n')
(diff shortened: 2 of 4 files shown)


CI/scripts/test.sh
```diff
@@ -202,22 +202,39 @@ function test-mtls-ssl-client-server-comm-pytest() {
...
 
+    # ----
     # Basic test to exercise core certificate verification v/s root-CA-cert
     pytest --capture=tee-sys -v test_client_server_mtls.py -k test_verify_certs_versus_root_cert
 
+    # ----
     # Exercise client-server communication with root-CA-signed certificates
-    pytest --capture=tee-sys -v test_client_server_mtls.py -k test_server_process_with_mtls_root_signed_cert &
+    pytest --capture=tee-sys -v test_client_server_mtls.py \
+            -k test_server_process_with_mtls_root_signed_cert_and_pvt_key_file &
```
gapisback, line 220: No change in test logic. Just renamed this to clarify that this test case will use (default) private-key file written to some known location in provisioning dir.

CI/scripts/test.sh
```diff
@@ -202,22 +202,39 @@ function test-mtls-ssl-client-server-comm-pytest() {
...
-    pytest --capture=tee-sys -v test_client_server_mtls.py -k test_client_app_with_mtls_root_signed_cert
+    pytest --capture=tee-sys -v test_client_server_mtls.py \
+            -k test_client_app_with_mtls_root_signed_cert_and_pvt_key_file
+
+    # ----
+    # Exercise client-server communication with root-CA-signed certificates
+    # using private-key written-to and read-from a temp-file, for certificate
+    # verification.
+    pytest --capture=tee-sys -v test_client_server_mtls.py \
+            -k test_server_process_with_mtls_root_signed_cert_using_temp_pvt_key_file &
```
gapisback, line 232: New test case which will write out private-key data to a named temporary file, in an anonymous (but named) temp-dir.

CI/scripts/test.sh
```diff
@@ -202,22 +202,39 @@ function test-mtls-ssl-client-server-comm-pytest() {
...
+
+    sleep 5
+
+    pytest --capture=tee-sys -v test_client_server_mtls.py \
+            -k test_client_app_with_mtls_root_signed_cert_using_temp_pvt_key_file
```
gapisback, line 237: Matching client-side-invocation case for above server-invocation test case.

CI/scripts/test.sh
```diff
@@ -480,8 +497,36 @@ function test-run_example-simple_app_python() {
...
+    # This invokes the 'run-app-as-server' operation in example_app.py
+    # which does a warm_restart() to re-load trust data from the policy-store.
+    ./run_example.sh simple_app_python run_app_as_server_offers_trusted_service
+
+    # This invokes the 'run-app-as-client' operation in example_app.py
+    # which does a warm_restart() to re-load trust data from the policy-store.
+    # There is assert code way-deep-down in example_app.py that verifies that
+    # the client gets an expected 'Hello' message from the server, through this
+    # exchange.
+    ./run_example.sh simple_app_python run_app_as_client_make_trusted_request
```
gapisback, line 528: Comments should be self-explanatory.

sample_apps/simple_app_python/example_app.py
```diff
@@ -163,12 +164,6 @@ def do_main(args) -> bool:
             print(fnl(), 'certify_me() failed\n')
             sys.exit(1)
 
-        # Server will persist required files upon certification.
-        if appln_data_dir.endswith(SERVER_APP_DATA):
-            whoami = 'server'
-            # Persist cert-and-key files in the provisioning dir.
-            write_cert_pvt_key_to_files(cctm, whoami, PROVISIONING_DIR, print_all)
```
gapisback, line 170: Reworked this logic, so that we now can show that client & server can be re-started. And that `warm_restart()` correctly re-installs trust data from policy-store.

Hence, we no longer write-out certificate data upon initial `get-certified` certification. This will be re-written-out each time server & / or client-app is restarted.

sample_apps/simple_app_python/example_app.py
```diff
@@ -178,8 +173,9 @@ def do_main(args) -> bool:
             print(fnl(), 'warm_restart() failed\n')
             sys.exit(1)
 
-        result = server_dispatch(PROVISIONING_DIR,
-                                 server_app_host, server_app_port)
+        write_certficates_to_file(cctm, 'server', PROVISIONING_DIR, print_all)
```
gapisback, line 176: Only certificate data will be written out to known-location in provisioning dir by this method.

sample_apps/simple_app_python/example_app.py
```diff
@@ -202,55 +198,52 @@ def do_main(args) -> bool:
...
+                  file(s) has been persisted earlier
+      server_app_host
+      server_app_port
+                - Server app's hostname and port number
+    """
+    print('\n**** ', fnl(), 'Start server process ...\n')
 
-    ca_root_outfilename = os.path.join(data_dir, ROOT_CA_POLICY_CERT)
-    context.load_verify_locations(cafile = ca_root_outfilename)
+    ssl_ctxt = setup_server_ssl_context(cctm, data_dir, print_all)
```
gapisback, line 246: This modularization is modeled on what was done for the pytest case, `test_server_process_with_mtls_root_signed_cert_using_temp_pvt_key_file()`, which also calls a similar `setup_server_ssl_context()` method.

sample_apps/simple_app_python/example_app.py
```diff
@@ -277,28 +270,71 @@ def server_dispatch(data_dir, server_app_host, server_app_port):
...
     """
-    print(fnl(), 'Start client application process ...\n')
+    Setup a SSL context for server-process, verifying certificates v/s
+    root certificate issued by CA.
 
-    # Client needs to authenticate the server; hence 'Purpose.SERVER_AUTH'
-    context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
+    **** NOTE: **** SSL interfaces need to read the private key from a file.
+
+    So, this method invokes a Certifier trust manager's method to persist
```
gapisback, line 280: This setup method is where we persist private-key data from trust_manager to [a] named temporary file, in an anonymous-but-named temp-dir.

sample_apps/simple_app_python/example_app.py
```diff
@@ -277,28 +270,71 @@ def server_dispatch(data_dir, server_app_host, server_app_port):
...
-    context.load_cert_chain(certfile = cert_outfilename, keyfile = key_outfilename)
+
+    with TemporaryDirectory() as tempdir:
+        with NamedTemporaryFile('wt', dir=tempdir, delete=False) as temp_keyf:
+            temp_keyf.close()
+
+            dump_private_key_to_file(cctm, temp_keyf.name, 'server', print_all)
+
+            context.load_cert_chain(certfile = cert_outfilename,
+                                    keyfile = temp_keyf.name)
```
gapisback, line 308: The security-"hole" is between lines L305 and L308.

sample_apps/simple_app_python/example_app.py
```diff
@@ -318,7 +354,52 @@ def client_dispatch(data_dir, server_app_host, server_app_port):
...
+    cert_outfilename = os.path.join(data_dir, APPLN_CERT_SIGNED_BY_ROOT_CA)
+
+    with TemporaryDirectory() as tempdir:
+        with NamedTemporaryFile('wt', dir=tempdir, delete=False) as temp_keyf:
+            temp_keyf.close()
+
+            dump_private_key_to_file(cctm, temp_keyf.name, 'client', print_all)
+
+            context.load_cert_chain(certfile = cert_outfilename,
+                                    keyfile = temp_keyf.name)
```
gapisback, line 394: Same pattern of new code flow & refactoring as was done on server-side.

sample_apps/simple_app_python/example_app.py
```diff
@@ -366,10 +447,50 @@ def parseargs(args):
     # ======================================================
     parser  = argparse.ArgumentParser(description='Certifier Framework: Simple App',
                                       formatter_class=argparse.RawDescriptionHelpFormatter,
-                                      epilog=r'''Examples:
-- Basic usage:
+                                      epilog=f'''Examples:
+
+- Run App-as-Server talk to Certifier Service:
```
gapisback, line 452: Added elaborate usage examples under `--help` output.

tests/README.md
```diff
@@ -34,10 +34,37 @@ Usage: test.sh [--help | --list] [ --from-test <test-function-name> ] [test_all]
     This will resume test execution from the `test-build-and-install-sev-snp-simulator`
     test to run the remaining tests in the order reported by the `--list` argument.
 
+## Executing test targets: `--list` output
```
gapisback, line 37: Elaborated usage of test-targets in `test.sh`. 

See this [tests/README](https://github.com/vmware-research/certifier-framework-for-confidential-computing/blob/agurajada/214-py-private-key/tests/README.md) or better rendition.

tests/pytests/test_client_server_mtls.py
```diff
@@ -17,43 +17,17 @@
 Basic pytests to exercise stripped-down Client-Server SSL Socket
 communication using mTLS.
 
-# pylint: disable=line-too-long
```
gapisback, line 20: Relocated to the bottom of the file, to reduce distraction upon initial reading ...

tests/pytests/test_client_server_mtls.py
```diff
@@ -197,17 +171,134 @@ def test_verify_certs_versus_root_cert():
...
+# ##############################################################################
+# To see output, run: pytest --capture=tee-sys -v
+def test_client_app_with_mtls_root_signed_cert_and_pvt_key_file():
+    """
+    Run client-application that sends message via secure SSL channel.
+    This exercises a client process talking to a server whose cert
+    was generated and signed by a root CA.
+    """
+    ssl_ctxt = setup_client_ssl_context()
+    setup_client_socket_comm_with_server(ssl_ctxt)
```
gapisback, line 191: Modularized inline code into sub-methods ... for code re-use.

sample_apps/simple_app_python/example_app.py
```diff
@@ -277,28 +270,71 @@ def server_dispatch(data_dir, server_app_host, server_app_port):
...
+    context = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
     context.verify_mode = ssl.CERT_REQUIRED
 
     cert_outfilename = os.path.join(data_dir, APPLN_CERT_SIGNED_BY_ROOT_CA)
-    key_outfilename  = os.path.join(data_dir, APPLN_KEYF_SIGNED_BY_ROOT_CA)
-    context.load_cert_chain(certfile = cert_outfilename, keyfile = key_outfilename)
+
+    with TemporaryDirectory() as tempdir:
+        with NamedTemporaryFile('wt', dir=tempdir, delete=False) as temp_keyf:
+            temp_keyf.close()
```
rgerganov, line 303: why do we have to close the file here?

sample_apps/simple_app_python/example_app.py
```diff
@@ -38,6 +39,7 @@
 
 # ------------------------------------------------------------------------------
 # Script global symbols to app-data dirs.
+CLIENT_APP_DATA='app1_data'
```
rgerganov, line 42: I think I said this in the previous PR but I will reiterate here -- this naming is confusing and doesn't follow the C++ example. Why do we need two data dirs? The example app should have a single data dir where the state is persisted. The name of this dir should be passed as command line argument (it's ok to have a default name). Then we can start a server instance of this example passing `app1_data` (or whatever) as data dir and another instance passing `app2_data` as data dir.

Having `PROVISIONING_DIR` also doesn't make a lot of sense. Why not simply store the app cert and the root cert in the data dir? The app cert **belongs** to the app and it make sense to be in its data dir.

jlmucb (approved):

sample_apps/simple_app_python/example_app.py
gapisback: Actually you are right, this is not really needed. Took it out and the test still works.

sample_apps/simple_app_python/example_app.py
gapisback: @rgerganov : I think there is some terminology confusion going on here.

Re: '_this naming is confusing and doesn't follow the C++ example._' - Not exactly right. yes, it's true that in the C++ sample program only one data-dir variable is declared. But if you see the invocation from `run_example.sh`, for instance for simple_app, you will see this program is invoked with diff `--data-dir` for client & server.

Here is the output:

```
Running dry-run run_app_by_name_as_server_talk_to_Cert_Service:

/home/agurajada/Code/certifier-framework-for-confidential-computing/sample_apps/simple_app/example_app.exe --data_dir=./app2_data/ --operation=cold-init --measurement_file=example_app.measurement --policy_store_file=policy_store --print_all=true
```

And for the client-app:

```
Running dry-run run_app_by_name_as_client_talk_to_Cert_Service

/home/agurajada/Code/certifier-framework-for-confidential-computing/sample_apps/simple_app/example_app.exe --data_dir=./app1_data/ --operation=cold-init --measurement_file=example_app.measurement --policy_store_file=policy_store --print_all=true
```

So we are doing what you say in your next comment; i.e. _Then we can start a server instance of this example passing app1_data (or whatever) as data dir and another instance passing app2_data as data dir._

---

We need two diff data_dirs, one each for the client & server, because each instance of the app in this client- / server-mode wants to store its trust_data in its own dedicated policy_store file on disk.

That's how the C++ apps work and I have continued the same model for the Python simple_app.

gapisback: About the provisioning dir - this is used by the driver script to generate policy-key and related files in one common dir which are then disbursed to the client- and server's app-data-dirs.

---

I think the Python simple_app is doing everything that the C++ app is also doing.

All you are seeing with these lines of code:

```
 42 CLIENT_APP_DATA='app1_data'
 43 SERVER_APP_DATA='app2_data'
```

is that these symbols are now used in this commit to generate `--help` messages, documenting the ways this program can be invoked. 

See the new messages added from L466 onwards.

Hope these explanations clarify the code-situation and your concerns are addressed.

rgerganov: Thanks, somehow I missed that `CLIENT_APP_DATA` and `SERVER_APP_DATA` are used only in the help message. I'd suggest to move them in the `parseargs()` function instead of having them in the global namespace.

I still don't get reason for having `PROVISIONING_DIR` though. Why not store the app cert and the root cert in the app's data dir? Right now you are passing this as `data_dir` argument to `write_certificates_to_file()` and I wonder why not using the data dir specified by the user. Why using hardcoded directory name (`./provisioning`) for this?

sample_apps/simple_app_python/example_app.py
rgerganov: There is a typo in the function name: `write_certficates_to_file -> write_certificates_to_file`

sample_apps/simple_app_python/example_app.py
gapisback: You are right, @rgerganov -- This is left-over cruft from my dev efforts on this sample program.

Indeed, `run_example.sh` disburses the generated binary cert / key-files from `provisioning` to app-specific data-dirs. So, yes, you are right; I should no longer need to rely on provisioning dir.

Cleaned up the program to address both your comments.

Thanks, for a careful review on this point ... and also for spotting the typo in `write_certficates_to_file()`.

gapisback:
@rgerganov - I cleaned up the PR based on your last set of comments. If you can give this a quick look, and approve, I'd like to push this to /main and close out this thread.

rgerganov (approved):
Thanks for addressing my comments

yelvmw (approved):
LGTM. Big thanks to both Aditya and Rado :-)

PR #222, opened 2023-11-03
(#215) Prevent rm_non_git_files from rm'ing user's config/env-files

gapisback:
This commit attempts to improve the behaviour of the rm_non_git_files sub-command of run_example.sh . The objective of this command is to clean-up build-outputs and other [stale] artifacts that may have been produced by a prior execution of this script. We may be, inadvertently, also deleting user's config-/env-files that reside in the project's sub-tree. This commit changes the behaviour of rm_non_git_files to now only delete the files registered in the .gitignore list.

Also, now "--dry-run rm_non_git_files" will only print an info message without actually deleting all affected files.

---
NOTE: @rgerganov - let's discuss this solution.

As mentioned in [this comment](https://github.com/vmware-research/certifier-framework-for-confidential-computing/issues/215#issuecomment-1792701926) for issue #215 , I am trying to clean-up the behaviour as best possible.

I'm not entirely sure whether driving off of `.gitignore` is risk-free or acceptable. There seem to be ways to have this `.gitignore` file in top-level root-dir and other sub-dirs. 

I'm not entirely sure if relying on output from `git ls-files . --exclude-standard --others --ignored` is sufficient and complete. But this does seem to work [well-enough] for normal usages in dev-env and thru the driver shell script.

An alternative would be to `find` files that need deleting (i.e. `*.o, *.so, *.exe, *.bin` ... and so on) and do a file-specific delete w/o relying on `git ls-files` output & behaviour. But this alternative may need to be continually updated if new artifacts are generated.

Another point is that the workflow steps do issue `make clean` in various places, so Makefile rules _should_ be cleaning stuff up. What remains is cleaning up the generated `*.bin` files that the workflow steps and this `run_example.sh` script stashes off in diff locations. `make clean` does not know about those files and will not `rm` them.

I'm not entirely happy with the current solution, but this may be good-enough. Let me know if this addresses your top-level concern with the old behaviour, that of possibly deleting user's config- / env-files that may be saved off in the projects' root dir.

### DIFF
--- a/sample_apps/run_example.sh
+++ b/sample_apps/run_example.sh
@@ -194,7 +194,7 @@ function usage() {
     echo "  To run an individual step of the ${SampleAppName}         : ./${Me} ${SampleAppName} <step name>"
     echo " "
     # Indentation to align with previous output lines.
-    echo "  Cleanup stale artifacts from build area             : ./${Me} rm_non_git_files"
+    echo "  Cleanup artifacts and stale outputs from build area : ./${Me} rm_non_git_files"
     echo "  Show the environment used to build-and-run a program: ./${Me} ${SampleAppName} show_env"
 
     local test_app="simple_app_under_oe"
@@ -203,7 +203,7 @@ function usage() {
     echo "  Setup and run the simple_app_under_oe program : ./${Me} --dry-run ${test_app}"
     echo "  Setup simple_app_under_oe program             : ./${Me} --dry-run ${test_app} setup"
     echo "  Run and test the simple_app_under_oe program  : ./${Me} --dry-run ${test_app} run_test"
-    echo "  Cleanup stale artifacts from build area       : ./${Me} --dry-run rm_non_git_files"
+    echo "  Cleanup artifacts and outputs from build area : ./${Me} --dry-run rm_non_git_files"
     echo " "
     echo "Options meant for developer usage:"
     echo "  --no-cleanup    : When errors occur, suppress executing cleanup.sh, which will kill active processes."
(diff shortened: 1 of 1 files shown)


sample_apps/run_example.sh
```diff
@@ -641,10 +641,16 @@ function run_popd() {
...
-# some prior steps.
+# some prior steps. This function will rm files that are listed in the
+# .gitignore file for this project. This way we won't delete any user's
+# config-/env-files saved for this project.
 # ###########################################################################
 function rm_non_git_files() {
     run_cmd
+    echo "${Me}: Delete all files not tracked by git that are also ignored."
+    if [ $DryRun -eq 1 ]; then
+        return
```
gapisback, line 652: Fixes an existing but that even in `--dry-run` mode this command was actually doing some work and deleting files. 

It should not do anything in dry-run mode.

sample_apps/run_example.sh
```diff
@@ -674,12 +680,6 @@ function rm_non_git_files() {
         Islet_lib_exists=1
     fi
 
-    echo "${Me}: Delete all files not tracked by git"
-
-    # shellcheck disable=SC2046
-    run_cmd rm -rf $(git ls-files . --exclude-standard --others)
```
gapisback, line 680: Delete this `rm`. Now we will only `rm` on L684 those files that are registered in `.gitignore` file.

rgerganov (approved):
I think this is fine as a stopgap solution. In the long term we should try to adopt a better build system (like CMake) which places all build artifacts in a separate directory.

Issue #154, opened 2023-07-19
Ensure that all shared-libs are measured and accounted-for in the measurement of Python applications.


gapisback:
PR #144 is introducing two shared libraries that are SWIG-generated Python wrappers around the Certifier Framework. These are:

1. `libcertifier_framework.so`
2. `libcertifier_tests.so`

(1) will be used to develop simple_app written in Python.

We need a mechanism to ensure that all such shared libraries used for Python (or other language) bindings are included when generating the measurement of Python-based applications using the Certifier Framework. Investigate and figure out how to do this work.

---
30.Oct.2023: Update from John:

1. Shared library that is in initram is already measured, so there should be nothing to be done for measurement.
2. If it comes from an untrusted disk, we have to do something.

gapisback:
HI, @jlmucb [ cc: @yelvmw, @rgerganov  ] -- before implementing anything, I want to run a proposal by you.

Currently, the `measurement_utility.exe` has been extended to include the contents of the generated `policy_key.py` file in the measurement. 

The usage of this tool is:

```
measurement_utility.exe     -input (input file) type: string default: "measurement_utility.exe"
                        -output (output file) type: string
                       -policy_key (policy key file, e.g., policy_key.py) type: string default: ""
```

The `-policy_key <policy key file>` argument was added in previous round of support for Python simple_app.

For the Python simple_app, at run-time, we need (and access) these two pieces of s/w lying in the build-area:

1. certifier_framework.py
2. libcertifier_framework.so

The Python bindings **will not work** if either of these is missing. Hence, I think these should be included in the sample programs "measurement".

We could extend the command-line flags to include a new `--cmd-line-flag` for each new file that has to be included in the measurement.

Alternatively, how about we just add one single `--other-files=<comma separated list of files to include>` argument.

Then, for Python apps, the tool will be invoked as:

```
measurement_utility.exe -input=example_app.py \
                    -other-files=policy_key.cc,certifier_framework.py,libcertifier_framework.so
```

The `-other-files` is a catch-all argument which can be used for other programs, applications that need more s/w artifacts to be included in the measurement. If any of the files listed by this argument are missing, then the utility will fail with an error.

This argument could become useful in future, too, as we discover the need to include more s/w components in the overall measurement of some running sample / application-program.

Rado -- when you ported the use of CF to Trezor demo-kit, did you find the need to include more s/w components in the measurement? Perhaps, we could learn from your experience while designing this command's interfaces.

[...]

PR #225, opened 2023-11-08
(#154) Extend measurement_utility to support --other_files argument

gapisback:
Python simple_app depends on two other artifacts:
  - certifier_framework.py - Python bindings to Certifier APIs
  - libcertifier_framework.so - Shared library

To be complete, include these two in the sample program's measurement. Extend measurement_utility to support a comma- separated list of files, specified by the --other_files argument. (This new arg replaces existing specialized --policy_key argument.)

Rework parsing this list to include the contents of these other files while computing the SHA-256 hash.

Update get_measurement_of_app_by_name() in run_example.sh to invoke this utility for the Python app case with this new argument, specifying the required additional 2 files.


sample_apps/run_example.sh
```diff
@@ -1043,13 +1043,14 @@ function get_measurement_of_app_by_name() {
     if [ "${SampleAppName}" = "application_service" ]; then
         measurement_file="app_service.measurement"
     elif [ "${SampleAppName}" = "simple_app_python" ]; then
-        policy_key_arg="--policy_key=../policy_key.py"
+        policy_key_arg="--other_files=../policy_key.py,${CERT_PROTO}/certifier_framework.py,${CERT_PROTO}/libcertifier_framework.so"
```
gapisback, line 1046: New arg `--other_files` to the measurement utility is applicable only for Python simple_app, here.

utilities/measurement_utility.cc
```diff
@@ -15,95 +15,127 @@
...
-             in_read,
-             policy_key_size,
-             policy_read);
+  // Read-in the other files to be measured, if provided
+  int other_files_read_size = 0;
+  if (other_files_total_size) {
+    to_hash += in_read;
+
+    // Read-in other files specified, after contents of input file read-in
+    for (int fctr = 0; fctr < (int)other_files_list.size(); fctr++) {
```
gapisback, line 99: Logic to iterate through parsed-list of files specified on the command-line.

utilities/measurement_utility.cc
```diff
@@ -114,16 +146,52 @@ int hash_utility(string &input, string &policy_key_file, string &output) {
...
+int parse_other_files_size(string &        other_files,
+                           vector<string> &other_files_list,
+                           vector<int> &   other_files_size) {
+  stringstream ss(other_files);
+
+  int other_files_total_size = 0;
+  while (ss.good()) {
+    string substr;
+    getline(ss, substr, ',');
+    int this_files_size = file_size(substr);
```
gapisback, line 174: `file_size()` does not return an error if file is not found. It returns -1, so check for that, and print an error here.

utilities/measurement_utility.cc
```diff
@@ -15,95 +15,127 @@
...
+                                                   other_files_list,
+                                                   other_files_size)
+                          : 0);
 
-  byte *       to_hash = (byte *)malloc(to_hash_size * sizeof(byte) + 1);
+  if (other_files_total_size < 0) {
+    printf("Error, reading one or more input files.\n");
+    return 1;
+  }
+  to_hash_size += (other_files_total_size + 1);
```
rgerganov, line 65: why do you add 1 here?

utilities/measurement_utility.cc
```diff
@@ -15,95 +15,127 @@
...
+
+  byte *       to_hash = (byte *)malloc(to_hash_size * sizeof(byte));
```
rgerganov, line 67: you don't need to load all files in memory to compute this hash -- you can load files one by one and update the hash

rgerganov (approved):
leaving some comments but up to you whether to address them now or later

utilities/measurement_utility.cc
gapisback: I just continued what was being done in the old pre-fix code, here:

```
-  byte *       to_hash = (byte *)malloc(to_hash_size * sizeof(byte) + 1);
```

I guess the old code was being careful to avoid off-by-one errors.

If this `+1` is not objectionably wrong ... I'd like to not mess with it, @rgerganov

utilities/measurement_utility.cc
jlmucb: If you load them one by one there is a TOCTOU vulnerability, I think.

jlmucb: Although this is not a big deal in a utility

jlmucb (approved):

utilities/measurement_utility.cc
gapisback: @rgerganov - I went back and visited the need for this `+ 1`. It's not needed. I took it out and confirmed that `../tests/measurement_utility_test.sh` still succeeds. Minor cleanup ...

utilities/measurement_utility.cc
gapisback: Hi, I went back and did some investigation on @rgerganov 's comment -- that would have simplified the code as we would not need to allocate a huge buffer to read-in contents of multiple files.

However, @jlmucb 's comment is helpful -- so I retained the code as it was submitted for review.

Still, before I got John's comment, I experimented whether Rado's suggestion would result in the same hash for multiple files, v/s running `digest_message()` across each file, and have it update the hash.

I maybe misunderstanding this approach, but take a look at the new test case added in `support_tests.cc/test_digest_multiple()` See my comments in that test case, which seems to indicate that as coded both these schemes are not similar. They result in different hash-digest values.

src/support_tests.cc
```diff
@@ -767,6 +770,135 @@ bool test_digest(bool print_all) {
...
+#define NUM_TEST_FILES_TO_DIGEST 4
+
+/*
+ * test_digest_multiple() - Similar to test_digest()
+ *
+ * Here we do a simple verification whether the hash-digest produced by
+ * digest_message() is the same if you read-in all input files once
+ * v/s if you read-on each file one-at-a-time and make multiple calls
+ * to digest_message(). Using digest_message(), which is a wrapper around
+ * underlying EVP_Digest*() interfaces, is not additive. This test checks
```
gapisback, line 782: @rgerganov & @jlmucb -- Please review this header block and my conclusion that "... is not additive", the way the `digest_message()` is coded.

If you both agree with this test-case, I'd like to retain it, as it closes out a doubt that someone else may have / raise in the future.

src/support_tests.cc
```diff
@@ -767,6 +770,135 @@ bool test_digest(bool print_all) {
...
+    if (!read_file(filename, &files_size[i], to_hash_single)) {
+      free(to_hash_single);
+      printf("Error: Cannot read file '%s'\n", filename);
+      return false;
+    }
+    if (print_all) {
+      printf("Read file: '%s', size=%d\n", filename, files_size[i]);
+    }
+    // Update digest-hash across each file
+    if (!digest_message(Digest_method_sha256,
```
gapisback, line 836: I noted that this wrapper interface is calling stuff like `EVP_DigestUpdate()`, followed by `EVP_DigestFinal_ex()`.

Seems like once you finalize the digest, then you have to re-start on the next call to this `digest_message()` with a new `EVP_MD_CTX *mdctx`.

I feel unless we re-work this `digest_message()` to -not- call finalize, so that it can be called iteratively on multiple files, we won't get the same resultant hash for all-files-in-one-go v/s each-file-digest method.

Let me know if I'm seeing things wrong ... or if this test-case is bogus and misadvertised.

utilities/measurement_utility.cc
```diff
@@ -15,95 +15,141 @@
...
+ * Compute the hash-digest for the input file and, optionally, include
+ * a list of other-files that need to be included in the measurement,
+ * specified by the other_files argument.
+ *
+ * NOTE: We make sure to read-in the contents of all files that go into the
+ * measurement ino one buffer. This slightly reduces the chance that while
+ * this utility is running, we don't run into the Time-of-Check to Time-of-Use,
+ * abbreviated as "TOCTOU", 'vulnerability'. That might happen if we read-in
+ * file-1 first, and while other files are being read, contents of file-1
+ * is changed. Second, and more importantly, we use the digest_message()
```
gapisback, line 53: If my assessment with the newly added test case is correct, then the 2nd point starting from "Second ..." is valid.

Otherwise, I'll have to withdraw the test case and rework this comment.

Please review this carefully and let's settle what the right thing to do is.

gapisback:
@jlmucb and @rgerganov -- I kicked this PR back to review to get your attention to a new test case I built,  `test_digest_multiple()`, to explore Rado's suggestion of updating the hash-digest one-file-at-a-time.

Please see my annotations in the [delta-commit](https://github.com/vmware-research/certifier-framework-for-confidential-computing/pull/225/commits/ab02176003ee0d04e20c1ee54e9f8e780d989382), leading you to new changes that need a re-think.

src/support_tests.cc
rgerganov: Correct. The current implementation of `digest_message` doesn't allow continuous hashing of multiple inputs.

utilities/measurement_utility.cc
rgerganov: I don't think that hashing files one by one creates any vulnerability here

rgerganov (approved):

utilities/measurement_utility.cc
jlmucb: yes

jlmucb (approved):

PR #227, opened 2023-11-08
Add IOCTL retries to SEV-SNP support

yelvmw:
SEV IOCTL fails when rate limiting kicks in. This patch adds the ioctl retries to handle the rate limiting case internally.

### DIFF
--- a/src/sev-snp/sev_support.cc
+++ b/src/sev-snp/sev_support.cc
@@ -48,7 +48,9 @@
 using namespace certifier::framework;
 using namespace certifier::utilities;
 
-#define SEV_GUEST_DEVICE "/dev/sev-guest"
+#define SEV_GUEST_DEVICE      "/dev/sev-guest"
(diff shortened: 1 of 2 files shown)


src/sev-snp/sev_support.cc
```diff
@@ -342,7 +364,7 @@ int sev_request_key(struct sev_key_options *options,
   }
 
   errno = 0;
-  rc = ioctl(fd, SNP_GET_DERIVED_KEY, &guest_req);
+  ioctl_with_retry(fd, SNP_GET_DERIVED_KEY, &rc, &guest_req);
```
rgerganov, line 367: you macro doesn't return -1 on error and directly returns `errno`; this might be a problem because on the next line we compare with -1 to detect errors

yelvmw: Will fix. Thanks!

jlmucb (approved):

rgerganov (approved):

PR #229, opened 2023-11-15
Sgx properties

jlmucb:
This adds platform property verification to Gramine enclaves.

This puts SGX on parity with SEV.


yelvmw:
It looks you decided to logically separate Gramine with and without platform evidence. This might be a simplification for now. But the result is that we either have to have platform properties in the policy or not. I don't think this is the right way to go. The choice should be handled given whether platform properties are present in the policies and this alone.

The current code will fail if I don't supply any platform properties. Additionally, I think we might have a bug in constructing the platform properties. If we remove a property from the policy (e.g., remove qeSVN, which should be valid), the verification should just treat it as if the specific property does not matter. However, the current code fails the verification. I'll dig deeper to see what went wrong.

yelvmw (approved):

Issue #174, opened 2023-08-07
Parametrize simple_app to specify combinations of public-/symmetric-key algorithms, for improved testing / stability.


gapisback:
Extend `simple_app` to specify public-key and symmetric-key algorithms, to exercise different combinations through this app. Keep defaults to "rsa-2048" and "aes-256-cbc-hmac-sha256", respectively.

----- This issue results from the following internal discussion. ----

A quick scan of all sample-apps shows that we are basically invoking just this pair of algorithms:
 
```
simple_app/example_app.cc:  string public_key_alg("rsa-2048");
simple_app/example_app.cc:  string symmetric_key_alg("aes-256-cbc-hmac-sha256");
 ```
From recent conversations, there are a collection of such algorithms in the Certifier code base.

Do we have anything else testing other public-key/symmetric-key algorithms’ implementation in our code base?

Currently, none of the sample-apps are exercising any other scheme. That seems like a testing hole to me.

In theory, one could parameterize these sample-apps to take these as parameters specified by `–public-key-alg` and `–symmetric-key-alg` arguments, with the default being their current behaviour.
 
Once we start exercising different pairs, we will know if the rest of the library / implementation is behaving correctly.
 
It’s fairly easy enough to parametrize simple_app and run through pair-wise diff algorithms to get improved coverage.

---
Reply from @jlmucb:

Yes, there are test for all the algorithms in support_tests.cc and the calls in cc_trust_data use only the tested algorithms.

None of the new code has anything to do with algorithm variety.  The new code works exactly like the old code in that respect.

Of course, you could paramaterize it in the tests by simply adding a flag in the app call instead of "hard wiring the algs."

PR #232, opened 2023-11-29
(#174) Initial support to parametrize public-key and symmetric-key algorithms.

gapisback:
This commit lays down the plumbing to parametrize the names of the public-key and symmetric-key algorithms that will be used by the simple_app. Currently, these are hard-coded to 'Enc_method_rsa_2048' and 'Enc_method_aes_256_cbc_hmac_sha256', respectively.

This commit now supports `--public_key_alg` and `--symmetric_key_alg` flags that can be supplied to the `simple_app/example_app.exe` program.

An initial list of `char *Enc_public_key_algorithms[]` and `char *Enc_symmetric_key_algorithms[]` has been created in `src/certifier_algorithms.cc`. The idea is that we can run test-executions choosing pairs of these algorithms to drive the simple_app.

## High-level changes:

- Support --public_key_alg and --symmetric_key_alg flags which can be supplied [only] to the simple_app/example_app.exe program.

- An initial list of char *Enc_public_key_algorithms[] and
  char *Enc_symmetric_key_algorithms[] has been created in src/certifier_algorithms.cc

- run_example.sh:
  - Refactor methods implementing simple_app_python tests into their own functions as argument-parsing is different for C++  v/s Python apps. (New args added with this work are n/a for the Python simple_app.)

  - Add new test-method run_test-crypto_algorithms(), which will exercise the sub-steps needed to run simple_app with different  pairs of crypto algorithms.

- CI/test.sh, build.yml: Add test-simple_app-with-crypto_algorithms, which will invoke said test case. This invokes
   run_test-crypto_algorithms() named earlier.

   - Pulled-out testing of simple_app_python with 'warm-restart'  functionality into its own test-case invocation. 
   -   Adds test-simple_app_python-with-warm-restart().

- Cleaned-up some dependency of 'setup' step for simple_apps to reduce CI test run-times.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,9 +96,26 @@ jobs:
         ./CI/scripts/test.sh test-run_example-simple_app
 
     #! -------------------------------------------------------------------------
+    - name: test-simple_app-with-crypto_algorithms
+      run: |
+        #! ---------------------------------------------------------------------
+        # Set this env-var to indicate that we are running this test-case
+        # from CI, where previous test case has already run 'setup'.
+        # So, in this test case, we will not re-do 'setup'.
+        FROM_CI_BUILD_YML=1 ./CI/scripts/test.sh test-simple_app-with-crypto_algorithms
+
+    #! -------------------------------------------------------------------------
     - name: test-run_example-simple_app_python
       run: |
         ./CI/scripts/test.sh test-run_example-simple_app_python
+
+    #! -------------------------------------------------------------------------
+    - name: test-simple_app_python-with-warm-restart
+      run: |
+        # Set this env-var to indicate that we are running this test-case
+        # from CI, where previous test case has already run 'setup'.
+        # So, in this test case, we will not re-do 'setup'.
+        FROM_CI_BUILD_YML=1 ./CI/scripts/test.sh test-simple_app_python-with-warm-restart
 
     #! -------------------------------------------------------------------------
     - name: test-build-and-setup-App-Service-and-simple_app_under_app_service
--- a/CI/scripts/test.sh
+++ b/CI/scripts/test.sh
@@ -24,6 +24,16 @@ if [ "${NumMakeThreads}" -gt 8 ]; then NumMakeThreads=8; fi
 # 'Globals' to track which test function is / was executing ...
 This_fn=""
 
+# -------------------------------------------------------------------------
+# Symbol to track dependency of test-simple_app_python-with-warm-restart
+# test-case on 'setup' done for simple_app_python. In CI, these two tests
+# are run back-to-back, so this dependency is assured. If the user runs
+# this test-case manually, we track this dependency via this global symbol
+# -------------------------------------------------------------------------
+Simple_app_python_setup_done="${FROM_CI_BUILD_YML:-0}"
+Simple_app_setup_done="${FROM_CI_BUILD_YML:-0}"
+
+# Symbol to track dependency of test-simple_app_python-with-warm-restart
 # ###########################################################################
 # Set trap handlers for all errors. Needs -E (-o errtrace): Ensures that ERR
 # traps (see below) get inherited by functions, command substitutions, and
@@ -59,7 +69,9 @@ TestList=( "test-core-certifier-programs"
            "test-run_example-help-list-args"
            "test-run_example-dry-run"
            "test-run_example-simple_app"
+           "test-simple_app-with-crypto_algorithms"
            "test-run_example-simple_app_python"
+           "test-simple_app_python-with-warm-restart"
            "test-build-and-setup-App-Service-and-simple_app_under_app_service"
            "test-run_example-multidomain_simple_app"
            "test-build-and-install-sev-snp-simulator"
@@ -431,9 +443,12 @@ function test-run_example-simple_app() {
 
     ./run_example.sh simple_app
 
+    # Just exec app exe to produce usage info
+    ./simple_app/example_app.exe
+
     # cp'over some certificate / policy / measurement bin files that were provisioned
     # for running the simple_app. The test cases in below test need these files to
-    # invoke (& verify) the steps under 'get-certified' target.
+    # invoke (& verify) the steps under the 'get-certified' target.
     pytest_data_dir="../tests/pytests/data/"
 
     for binfile in attest_key_file.bin \
@@ -451,9 +466,9 @@ function test-run_example-simple_app() {
     NO_ENABLE_SEV=1 make -f certifier.mak --always-make -j${NumMakeThreads} sharedlib
     popd > /dev/null 2>&1
 
-    # Re-start Certifier Service as script, above, would have shut it down
     ./cleanup.sh
 
+    # Re-start Certifier Service as script, above, would have shut it down
     ./run_example.sh simple_app start_certifier_service
 
     pgrep simple
@@ -468,6 +483,25 @@ function test-run_example-simple_app() {
 
     popd > /dev/null 2>&1
 
+    ./cleanup.sh
+
+    popd > /dev/null 2>&1
+}
+
+# #############################################################################
+function test-simple_app-with-crypto_algorithms() {
+    echo "******************************************************************"
+    echo "* Test: Execute script to compile, build and run simple_app."
+    echo "******************************************************************"
+    echo " "
+    pushd ./sample_apps > /dev/null 2>&1
+
+    ./cleanup.sh
+
+    if [ "${Simple_app_setup_done}" = 0 ]; then
+        ./run_example.sh simple_app setup
+    fi
+    ./run_example.sh simple_app run_test-crypto_algorithms
     ./cleanup.sh
 
     popd > /dev/null 2>&1
@@ -504,24 +538,45 @@ function test-run_example-simple_app_python() {
     ./run_example.sh simple_app_python setup
     ./run_example.sh simple_app_python run_test
 
-    # ---- Variation -----------------------------------------------------------
-    # Test the scenario that once certified, the server-process and client-app
-    # can be re-started multiple times, without need for the Certifier Service.
-    # Re-starting the server-process will re-load the trust data from the
-    # policy-store. It will no longer need to contact the Certifier Service
-    # to get re-certified.
-    # Then, you can re-start the client-app, to talk to this now-certified
-    # server-process directly through secure SSL channel.
-    # The following sequence of steps validate this workflow and verify that
-    # the client and server can communicate through this secure channel.
-    # --------------------------------------------------------------------------
+    ./cleanup.sh
+    popd > /dev/null 2>&1
+}
+
+# #############################################################################
+# Test the scenario that once certified, the server-process and client-app
+# can be re-started multiple times, without need for the Certifier Service.
+# Re-starting the server-process will re-load the trust data from the
+# policy-store. It will no longer need to contact the Certifier Service
+# to get re-certified.
+#
+# Then, you can re-start the client-app, to talk to this now-certified
+# server-process directly through secure SSL channel.
+# The following sequence of steps validate this workflow and verify that
+# the client and server can communicate through this secure channel.
+# #############################################################################
+function test-simple_app_python-with-warm-restart() {
     echo " "
     echo "*******************************************************************"
     echo "* Test Variation: Re-start server-process and client-app after shutting down the Certifier Service ..."
     echo "*******************************************************************"
     echo " "
 
-    ./cleanup.sh
+    pushd ./sample_apps > /dev/null 2>&1
+
+    # If this test case is being run on its own (outside of CI), we have to
+    # first go through setup and then run the test. This will get the
+    # server/client-apps 'certified'. After that, the meat of this test-case;
+    # i.e., re-run with warm-restart can be run.
+    set -x
+    if [ "${Simple_app_python_setup_done}" = 0 ]; then
+        ./run_example.sh rm_non_git_files
+        ./run_example.sh simple_app_python setup
+        ./run_example.sh simple_app_python run_test
+
+        # This will stop the Certifier Service.
+        ./cleanup.sh
+    fi
+    set +x
 
     # This invokes the 'run-app-as-server' operation in example_app.py
     # which does a warm_restart() to re-load trust data from the policy-store.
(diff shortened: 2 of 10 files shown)


src/certifier_algorithms.cc
```diff
@@ -49,15 +50,30 @@ const char * Enc_method_rsa_4096                  = "rsa-4096";
 const char * Enc_method_rsa_4096_private          = "rsa-4096-private";
 const char * Enc_method_rsa_4096_public           = "rsa-4096-public";
 const char * Enc_method_rsa_4096_sha384_pkcs_sign = "rsa-4096-sha384-pkcs-sign";
- 
+
+const char *Enc_public_key_algorithms[] =
```
gapisback, line 54: HI, @jlmucb - I need your help in defining the list of public-key algorithms in this array.

And the list of symmetric-key algorithms defined on L62.

I took an initial stab, but seems like from the full list of `Enc_method_*` algorithm names, we should be able to separate out public-key and symmetric-key algorithms into two buckets.

Can you put down the list(s), so I can continue code-dev on this PR to address issue #174?

.github/workflows/build.yml
```diff
@@ -95,11 +95,28 @@ jobs:
...
         ./CI/scripts/test.sh test-run_example-simple_app
 
+    #! -------------------------------------------------------------------------
+    - name: test-simple_app-with-crypto_algorithms
+      run: |
+        #! ---------------------------------------------------------------------
+        # Set this env-var to indicate that we are running this test-case
+        # from CI, where previous test case has already run 'setup'.
+        # So, in this test case, we will not re-do 'setup'.
+        FROM_CI_BUILD_YML=1 ./CI/scripts/test.sh test-simple_app-with-crypto_algorithms
```
gapisback, line 105: Exercise the testing hook provided by `run_example.sh` to run `simple_app` with diff pairs of named crypto algorithms.

.github/workflows/build.yml
```diff
@@ -95,11 +95,28 @@ jobs:
...
       run: |
         ./CI/scripts/test.sh test-run_example-simple_app_python
 
+    #! -------------------------------------------------------------------------
+    - name: test-simple_app_python-with-warm-restart
+      run: |
+        # Set this env-var to indicate that we are running this test-case
+        # from CI, where previous test case has already run 'setup'.
+        # So, in this test case, we will not re-do 'setup'.
+        FROM_CI_BUILD_YML=1 ./CI/scripts/test.sh test-simple_app_python-with-warm-restart
```
gapisback, line 118: Pulled out of previous test case on L110 to run as a stand-alone test-case. This externalizes the testing work being done specifically to test-out `warm-restart` capability.

CI/scripts/test.sh
```diff
@@ -473,6 +488,25 @@ function test-run_example-simple_app() {
...
+    echo "******************************************************************"
+    echo " "
+    pushd ./sample_apps > /dev/null 2>&1
+
+    ./cleanup.sh
+
+    if [ "${Simple_app_setup_done}" = 0 ]; then
+        ./run_example.sh simple_app setup
+    fi
+    ./run_example.sh simple_app run_test-crypto_algorithms
```
gapisback, line 504: Just a driver method in `test.sh` to call into the testing method in the sub-script.

CI/scripts/test.sh
```diff
@@ -504,24 +538,45 @@ function test-run_example-simple_app_python() {
...
+# Re-starting the server-process will re-load the trust data from the
+# policy-store. It will no longer need to contact the Certifier Service
+# to get re-certified.
+#
+# Then, you can re-start the client-app, to talk to this now-certified
+# server-process directly through secure SSL channel.
+# The following sequence of steps validate this workflow and verify that
+# the client and server can communicate through this secure channel.
+# #############################################################################
+function test-simple_app_python-with-warm-restart() {
```
gapisback, line 557: Externalized inline testing code into standalone test method.

sample_apps/common/example_app.cc
```diff
@@ -283,28 +288,70 @@ int main(int an, char **av) {
...
+#ifdef SIMPLE_APP
+    printf("\n\
+                  --platform_file_name=platform-cert-bin-file-name        ; %s\n\
+                  --platform_attest_endorsement=endorsement-bin-file-name ; %s\n\
+                  --measurement_file=measurement-bin-file-name            ; %s\n\
+                  --attest_key_file=attest-key-bin-file-name              ; %s\n",
+                  FLAGS_platform_file_name.c_str(),
+                  FLAGS_platform_attest_endorsement.c_str(),
+                  FLAGS_measurement_file.c_str(),
+                  FLAGS_attest_key_file.c_str());
```
gapisback, line 319: Above block of diffs improves the output from `--help` to be a bit more readable.

gapisback: Output now looks like this:

```
cc-sdb-vm:[237] $ ./example_app.exe
                                                                            (Defaults)
./example_app.exe --operation=<op>                                        ; (See below)
                  --policy_host=policy-host-address                       ; localhost
                  --policy_port=policy-host-port                          ; 8123
                  --server_app_host=my-server-host-address                ; localhost
                  --server_app_port=my-server-port-number                 ; 8124
                  --data_dir=-directory-for-app-data                      ; ./app1_data/
                  --policy_cert_file=self-signed-policy-cert-file-name    ;
                  --policy_store_file=policy-store-file-name              ; store.bin
                  --print_all=true|false
                  --platform_file_name=platform-cert-bin-file-name        ; platform_file.bin
                  --platform_attest_endorsement=endorsement-bin-file-name ; platform_attest_endorsement.bin
                  --measurement_file=measurement-bin-file-name            ; example_app.measurement
                  --attest_key_file=attest-key-bin-file-name              ; attest_key_file.bin


Operations are: cold-init, get-certified, run-app-as-client, run-app-as-server
```

sample_apps/run_example.sh
```diff
@@ -515,6 +524,13 @@ function is_valid_step() {
         return
     fi
 
+    # Special-hook to run simple_app to exercise diff crypto algorithms
+    if [ "${SampleAppName}" == "simple_app" ] \
+    && [ "${fn}" == "run_test-crypto_algorithms" ]; then
```
gapisback, line 529: We now will allow this as a valid invocation: `$ ./run_example.sh simple_app run_test-crypto_algorithms`

sample_apps/run_example.sh
```diff
@@ -3078,6 +3194,52 @@ function run_test_simple_app_under_app_service() {
...
+
+    run_cmd
+
+    start_certifier_service
+
+    subcase=0
+    # These valid-lists are found in certifier_algorithms.c
+    # public-key algo 'ecc-384' is failing ... Comment out for now ...
+    # for public_key_algo in rsa-3072 rsa-4096 ecc-384; do
+    for public_key_algo in rsa-3072 rsa-4096; do
```
gapisback, line 3213: Attn: @jlmucb - We need to discuss and reconcile exactly what public-key algorithms are supported.

I thought I could list many from `certifier_algorithms.cc` as they are enumerated in the newly-added `Enc_public_key_algorithms[]` array.

But due to the checks in `src/cc_helpers.cc/generate_auth_key()`, only a few public-key algorithms are actually supported or can be used in this pair-wise algorithms testing scheme.

For instance, an attempt to use `ecc-384`, i.e., `Enc_method_ecc_384`, will fail due to the check here in the above mentioned routine:

```
1154 bool certifier::framework::cc_trust_manager::generate_auth_key(bool regen) {
[...]
1188   } else {
1189     printf("%s() error, line %d, Unsupported public key algorithm: '%s'\n",
1190            __func__,
1191            __LINE__,
1192            public_key_algorithm_.c_str());
1193     return false;
1194   }
```

sample_apps/run_example.sh
```diff
@@ -3078,6 +3194,52 @@ function run_test_simple_app_under_app_service() {
...
+        # for symmetric_key_algo in aes-256-cbc-hmac-sha256; do
+        symmetric_key_algo="aes-256-cbc-hmac-sha256"
```
gapisback, line 3215: Similarly, seems like we can only use this one named symmetric-key algorithm.

Otherwise, we trip-up in `src/certifier.cc`, here:

```
 901 bool certifier::framework::unprotect_blob(const string &enclave_type,
 902                                           int           size_protected_blob,
 903                                           byte *        protected_blob,
 904                                           key_message * key,
 905                                           int * size_of_unencrypted_data,
 906                                           byte *unencrypted_data) {
 907
[...]
 958   if (key->key_type() != Enc_method_aes_256_cbc_hmac_sha256) {
 959     printf("%s() error, line %d, unprotect_blob, unsupported encryption "
 960            "scheme: '%s'\n",
 961            __func__,
 962            __LINE__,
 963            key->key_type().c_str());
 964     return false;
 965   }
```

What is the support for other crypto algorithms listed as `private` in this file:

```
cc-sdb-vm:[9] $ grep Enc_ certifier_algorithms.cc |  grep -i private
const char * Enc_method_ecc_256_private           = "ecc-256-private";
const char * Enc_method_ecc_384_private           = "ecc-384-private";
const char * Enc_method_rsa_1024_private          = "rsa-1024-private";
const char * Enc_method_rsa_2048_private          = "rsa-2048-private";
const char * Enc_method_rsa_3072_private          = "rsa-3072-private";
const char * Enc_method_rsa_4096_private          = "rsa-4096-private";
```

We need to complete the list of symmetric-key algorithms supported to finish this testability matrix.

include/certifier_algorithms.h
```diff
@@ -69,7 +69,7 @@ extern const char *Integrity_method_aes_256_gcm;
 extern const char *Integrity_method_hmac_sha256;
 
 extern const char *Enc_public_key_algorithms[];
-extern const char *Enc_symmetric_key_algorithms[];
+extern const char *Enc_authenticated_symmetric_key_algorithms[];
```
gapisback, line 72: Renamed name of array of string constants, to explicitly call-out the 'Authenticated' part of these algorithms.

sample_apps/common/example_app.cc
```diff
@@ -55,10 +55,11 @@ DEFINE_string(platform_attest_endorsement,
               "platform endorsement of attest key");
 DEFINE_string(attest_key_file, "attest_key_file.bin", "attest key");
 DEFINE_string(measurement_file, "example_app.measurement", "measurement");
+
 DEFINE_string(public_key_alg, Enc_method_rsa_2048, "public key algorithm");
-DEFINE_string(symmetric_key_alg,
+DEFINE_string(auth_symmetric_key_alg,
```
gapisback, line 60: For brevity, the command-line flag to name these algorithms will be `--auth_symmetric_key_alg`

sample_apps/run_example.sh
```diff
@@ -3211,9 +3211,14 @@ function run_test-crypto_algorithms() {
...
     for public_key_algo in rsa-3072 rsa-4096; do
-        # for symmetric_key_algo in aes-256-cbc-hmac-sha256; do
-        symmetric_key_algo="aes-256-cbc-hmac-sha256"
-
+        # RESOLVE: Investigate & fix these limitations:
+        #  - generate_symmetric_key() does not support
+        #    Enc_method_aes_128_cbc_hmac_sha256; i.e., "aes-128-cbc-hmac-sha256";
+        # - unprotect_blob() only allows Enc_method_aes_256_cbc_hmac_sha256
+        # - These two should also be supported in loop below:
+        #   - aes-256-cbc-hmac-sha384, aes-256-gcm
```
gapisback, line 3219: @jlmucb -- We need to discuss whether the limitations in the fns-mentioned are valid. 

The 3 authenticated symmetric-key algorithms named above do not work when we attempt to run `simple_app` with the newly added  `--auth_symmetric_key_alg` flag naming these algorithms.

gapisback:
Hi, @jlmucb - I did a minor rework of this change-set subsequent to our discussion this morning. 

See the diffs applied in this commit.

There are still some things that do not work:

1. Using `ecc-384` as a public-key algorithm name.
2. Using any one of these as an authenticated symmetric-key algorithm name: `aes-128-cbc-hmac-sha256`, or `aes-256-cbc-hmac-sha384`, `aes-256-gcm`. These run into limitations as documented in `run_example.sh` around L3214.

I'd like to get the bulk of the rest of this testing / refactoring machinery reviewed and merged.

Then, you / I can separately investigate why the choice of these diff crypto algorithms is not working.

src/certifier_algorithms.cc
```diff
@@ -49,15 +53,34 @@ const char * Enc_method_rsa_4096                  = "rsa-4096";
...
```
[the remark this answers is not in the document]
gapisback: @rgerganov - This is existing code I added some many cycles ago. See PR #176.

I had tried to use C++ `String` in that work but could not do that for the reasons documented in [this comment](https://github.com/vmware-research/certifier-framework-for-confidential-computing/pull/176#discussion_r1287808897).

I don't think it's great huge value now to move to `std::vector` to refactor this plumbing.

If there is a strong need to do so, it will be better to refactor that in a different change-set once this one goes in. (But I won't have time to follow-up on that.)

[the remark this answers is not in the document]
gapisback: @rgerganov - On a 2nd read, I think I misunderstood your comment.

I guess, yes, for the new static string array I _could try_ to see if `std::vector` will work.

Let me get all the comments from @yelvmw and you, and will attempt this rework in one pass at the end, while applying changes to address all review feedback.

yelvmw (approved):

src/certifier_algorithms.cc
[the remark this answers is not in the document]
gapisback: @rgerganov - I amended the definitions of these string arrays to be like so:

```
 57 std::vector<const char *> Enc_public_key_algorithms =
 58                     {   Enc_method_rsa_2048
 59                       , Enc_method_rsa_3072
 60                       , Enc_method_rsa_4096
 61                       , Enc_method_ecc_384
 62                     };
 63
 64 const int Num_public_key_algorithms = Enc_public_key_algorithms.size();
 65
 66 // Names of Authenticated symmetric-key algorithms
 67 std::vector<const char *> Enc_authenticated_symmetric_key_algorithms =
 68                     {   Enc_method_aes_128_cbc_hmac_sha256
 69                       , Enc_method_aes_256_cbc_hmac_sha256
 70                       , Enc_method_aes_256_cbc_hmac_sha384
 71                       , Enc_method_aes_256_gcm
 72                     };
 73
 74 const int Num_symmetric_key_algorithms = Enc_authenticated_symmetric_key_algorithms.size();
 75
```

This change is the only item you requested.

So, as we discussed this morning, I am going to squash all commits and will merge this change-set into `/main` once CI passes.

Issue #242, opened 2024-04-25
CI: make in ./sev-snp-simulator dir fails with fatal error: error: passing argument 1 of ‘class_create’ from incompatible pointer type


gapisback:
[CI Job](https://github.com/ccc-certifier-framework/certifier-framework-for-confidential-computing/actions/runs/8823860913/job/24225212066?pr=241) for PR #241 fails with following hard error:

```
make[1]: Leaving directory '/usr/src/linux-headers-6.5.0-1018-azure'
rm -f modules.order Module.symvers Module.markers *.o *.cmd sevnull.ko
rm -rf keys
make -C /lib/modules/6.5.0-1018-azure/build M=/home/runner/work/certifier-framework-for-confidential-computing/certifier-framework-for-confidential-computing/sev-snp-simulator modules
make[1]: Entering directory '/usr/src/linux-headers-6.5.0-1018-azure'
warning: the compiler differs from the one used to build the kernel
  The kernel was built by: x86_64-linux-gnu-gcc-11 (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
  You are using:           gcc-11 (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
  CC [M]  /home/runner/work/certifier-framework-for-confidential-computing/certifier-framework-for-confidential-computing/sev-snp-simulator/sevguest.o
In file included from ./include/linux/linkage.h:7,
                 from ./arch/x86/include/asm/cache.h:5,
                 from ./include/linux/cache.h:6,
                 from ./include/linux/time.h:5,
                 from ./include/linux/stat.h:19,
                 from ./include/linux/module.h:13,
                 from /home/runner/work/certifier-framework-for-confidential-computing/certifier-framework-for-confidential-computing/sev-snp-simulator/sevguest.c:14:
/home/runner/work/certifier-framework-for-confidential-computing/certifier-framework-for-confidential-computing/sev-snp-simulator/sevguest.c: In function ‘sev_guest_init_module’:
./include/linux/export.h:29:22: error: passing argument 1 of ‘class_create’ from incompatible pointer type [-Werror=incompatible-pointer-types]
   29 | #define THIS_MODULE (&__this_module)
      |                     ~^~~~~~~~~~~~~~~
      |                      |
      |                      struct module *
[...]

PR #243, opened 2024-04-25
Stabilize CI-build-and-test jobs on Ubuntu Linux 22.04

gapisback:
This commit attempts to keep the CI tests runing cleanly. 

Some tests  are failing due to different issues reported under issue #242 and #244. 

Those will need some minor corrections in build-steps.

For now, 5 test-cases from test.sh are commented out with a  Warning message.

 - Skip few simulated-SEV related build-and-test cases.
 - Skip ISLET-related build-and-test tests due to rustc version incompatibility.

Those will need to be re-enabled when above two issues are resolved, if possible.

(If further investigation shows that we cannot support Ubunutu 22.04, then some changes will need to be made in CI's build.yml to drop-down  to an older Ubuntu rev-level.)

### DIFF
--- a/sample_apps/run_example.sh
+++ b/sample_apps/run_example.sh
@@ -681,7 +681,9 @@ function rm_non_git_files() {
     fi
 
     local tmp_islet_dir="/tmp/islet-save"
-    mkdir "${tmp_islet_dir}"
+    if [ ! -d "${tmp_islet_dir}" ]; then
+        mkdir "${tmp_islet_dir}"
+    fi
     local Islet_remote_exists=0
     if [ -d "${ISLET_ROOT}/remote" ]; then
         set -x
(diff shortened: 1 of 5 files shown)


sample_apps/run_example.sh
```diff
@@ -681,7 +681,9 @@ function rm_non_git_files() {
     fi
 
     local tmp_islet_dir="/tmp/islet-save"
-    mkdir "${tmp_islet_dir}"
+    if [ ! -d "${tmp_islet_dir}" ]; then
+        mkdir "${tmp_islet_dir}"
```
gapisback, line 685: Minor unrelated fix. Re-runs of this script will fail if this dir exists.

sev-snp-simulator/Makefile
```diff
@@ -48,7 +48,7 @@ clean:
 	rm -f modules.order Module.symvers Module.markers *.o *.cmd sevnull.ko
 	rm -rf keys
 
-.PHNOY:
+.PHONY:
```
gapisback, line 51: Fix typo.

yelvmw (approved):
LGTM. Thanks!

PR #245, opened 2024-04-30
Fix to use the latest Islet SDK 

bitboom:
This PR addresses build errors in the Islet SDK. (Fixed https://github.com/ccc-certifier-framework/certifier-framework-for-confidential-computing/issues/244)

**Issues and Resolutions:**

1. **Issue:** Compatibility issues due to the upgrade of a third-party library used by the Islet SDK.
   **Fixing:** Upgraded to a compatible version and enforced explicit versioning to prevent automatic upgrades of the third-party library. (https://github.com/islet-project/islet/pull/315)

2. **Issue:** Upgrade of the MRENCLAVE value for the simulated version of Islet SDK targeting x86_64.
   **Fixing:** Updated to the upgraded value and added comments on how to obtain it. Although a shell script should ideally be written to fetch the actual value systematically, frequent changes are not anticipated, so only a description has been added for now.

### DIFF
--- a/sample_apps/run_example.sh
+++ b/sample_apps/run_example.sh
@@ -1108,7 +1108,7 @@ function get_measurement_of_trusted_simple_app_under_islet() {
     # get_measurement_of_app_by_name "islet_example_app.exe"
 
     run_cmd
-    local mrenclave="6190EB90B293886C172EC644DAFB7E33EE2CEA6541ABE15300D96380DF525BF9"
+    local mrenclave="580bd77074f789f34841ea9920579ff29a59b9452b606f73811132b31c689da9"
     local measurement_file="example_app.measurement"
 
     run_pushd "${PROV_DIR}"
--- a/sample_apps/run_example.sh
+++ b/sample_apps/run_example.sh
@@ -1108,6 +1108,10 @@ function get_measurement_of_trusted_simple_app_under_islet() {
     # get_measurement_of_app_by_name "islet_example_app.exe"
 
     run_cmd
+    # The following mrenclave value should hardly change as it is
+    # a simulated version targeting x86_64. (The actual version only works on aarch64.)
+    # If you want to manually obtain the mrenclave value, refer to the islet CLI at the link below.
+    # https://github.com/islet-project/islet/blob/main/examples/cross-platform-e2ee/README.md#2-configure-measurements-and-policies
     local mrenclave="580bd77074f789f34841ea9920579ff29a59b9452b606f73811132b31c689da9"
     local measurement_file="example_app.measurement"
 
--- a/third_party/islet/setup.sh
+++ b/third_party/islet/setup.sh
@@ -29,14 +29,15 @@ ISLET_INC="$HERE/include"
 ISLET_LIB="$HERE/lib"
 
 TARGET_HDR="$ISLET_SDK/include/islet.h"
-TARGET_LIB="$ISLET/out/x86_64-unknown-linux-gnu/debug/libislet_sdk.so"
+TARGET_LIB="$ISLET/out/x86_64-unknown-linux-gnu/debug/libislet_sdk.a"
 
 # Sync islet
 cd "$HERE"
-wget https://github.com/Samsung/islet/archive/refs/tags/certifier-v1.0.1-beta.tar.gz
-tar xf certifier-v1.0.1-beta.tar.gz
 rm -rf "$ISLET"
-mv islet-certifier-v1.0.1-beta "$ISLET"
+git clone https://github.com/islet-project/islet.git "$ISLET"
+cd "$ISLET"
+git submodule update --init --depth 1 $ISLET/third-party/ciborium
+git submodule update --init --depth 1 $ISLET/third-party/coset
 
 # Install rust to build ISLET SDK
 "$ISLET/scripts/deps/rust.sh"
--- a/CI/scripts/test.sh
+++ b/CI/scripts/test.sh
@@ -791,9 +791,6 @@ function test-ISLET-SDK-shim_test() {
     echo "* Download ISLET SDK, build the library and run shim_test"
     echo "**********************************************************"
     echo " "
-    echo "****** WARNING! Skipped due to open issue #242, Fails on Ubuntu 22.04.4"
-    echo " "
-    return
 
     pushd src/islet > /dev/null 2>&1
 
@@ -819,9 +816,6 @@ function test-run_example-simple_app_under_islet-using-shim() {
     echo "* Test: Execute script to compile, build and run simple_app_under_islet using shim"
     echo "***********************************************************************************"
     echo " "
-    echo "****** WARNING! Skipped due to open issue #242, Fails on Ubuntu 22.04.4"
-    echo " "
-    return
 
     pushd ./sample_apps > /dev/null 2>&1
 


gapisback:
Thanks for pulling this forward, @bitboom !

I see that CI tests ran clean with [your 1st commit](https://github.com/ccc-certifier-framework/certifier-framework-for-confidential-computing/pull/245/commits/9bfb680843be8b0585d22cd1cd0e37ad1a130381), but then when you tried to re-enable the failing (previously commented) ISLET build-and-test tests, they failed in [this build](https://github.com/ccc-certifier-framework/certifier-framework-for-confidential-computing/actions/runs/8889320667/job/24408630805).

I took a quick look and the failiure is here:

```
 /home/runner/work/certifier-framework-for-confidential-computing/certifier-framework-for-confidential-computing/sample_apps/simple_app_under_islet/islet_example_app.exe --data_dir=./app2_data/ --operation=get-certified --measurement_file=example_app.measurement --policy_store_file=policy_store '' '' --print_all=true
Simulated attestation operation on x86_64.
certify_domain():1735: enclave_type_ = 'islet-enclave', purpose_ = 'authentication'
islet_Attest(): rv=0
certify_domain() error, line: 2024, Certification failed, status='failed'

Response:
Status                 :  failed
++ cleanup
Requesting enclave     :  requesting-enclave
++ set +x
Providing  enclave     :  providing-enclave
certify_primary_domain() error, line 1477, can't certify primary domain
main() error, line 447, certification failed
run_example.sh: Failed command, "$@", at line 626 ./run_example.sh while executing function run_app_by_name_as_server_talk_to_Cert_Service
cleanup.sh: Cleanup stale processes (parent_pid=17455) ...
```

It was not immediately obvious to me if this is a failure due to (a) Islet-APIs changing or (b) if it's an error with driver `run_example.sh` script.

If it is (b), let me know. I can look at it later.

I was hoping to get the previous PR #243 integrated first, where failing tests are commented out. Then, you can layer your fix on top of that. Am waiting for Ye & you to sign-off on my PR which basically comments out failing tests.

bitboom:
> It was not immediately obvious to me if this is a failure due to (a) Islet-APIs changing or (b) if it's an error with driver `run_example.sh` script.

It is assumed that there might be a problem with the `openssl certificate` depending on the environment.

Because Islet API was successful on CI and my environment too.
> islet_Attest(): rv=0

The certificate error did not occur in my environment but occurred in the CI as you said.

```sh
run_example.sh: simple_app_under_islet: Running run_app_by_name_as_client_make_trusted_request

Running App as client
Client peer id is Measured-6190eb90b293886c172ec644dafb7e33ee2cea6541abe15300d96380df525bf9
SSL client read: Hi from your secret server
`

**Comment by gapisback** [comment]
Hi, @bitboom, 

Re: _It is assumed that there might be a problem with the openssl certificate depending on the environment._

... The CI `build.yml` is doing standard install of openssl libraries:

```
 17     - name: dependencies
 18       run: |
 19         sudo apt update -y
 20         sudo apt install -y clang-format-11 libgtest-dev libgflags-dev openssl libssl-dev protobuf-compiler proto    c-gen-go golang-go cmake
```

We will need to debug this in CI's env and on your machine, to compare the OpenSSL library versions.

That is a bit of painful effort.

What you will need to do is to edit out all the test-methods (after install) in `.github/workflows/build.yml` and just keep this these test-methods active. 

Check-in this to your dev-branch so that the CI-roundtrip is shortened to just this set of ISLET related failing test-methods.

```
155     #! -------------------------------------------------------------------------
156     - name: test-ISLET-SDK-shim_test
157       run: |
158         ./CI/scripts/test.sh test-ISLET-SDK-shim_test
159
160     #! -------------------------------------------------------------------------
161     - name: test-run_example-simple_app_under_islet-using-shim
```


And add some commands in the `name: dependencies` step to report the openssl version.

I don't know what / how to report the OpenSSL version string. But if you can add that, we can compare the version string that CI installs v/s what's on your dev-box.

Once we find out what the differences are ... we can tweak the `sudo apt install openssl` command to specify a particular version string.

I think this troubleshooting / debugging will be a slow exercise.

I'll leave it up to you to figure out if this is the best way to troubleshoot, or you have any better ideas to figure out what the OpenSSL version issue is (or might be).

File: sample_apps/run_example.sh [hunk 1]
```diff
@@ -681,9 +681,7 @@ function rm_non_git_files() {
     fi
 
     local tmp_islet_dir="/tmp/islet-save"
-    if [ ! -d "${tmp_islet_dir}" ]; then
-        mkdir "${tmp_islet_dir}"
-    fi
```
gapisback: Can you please retain this `if()` check I added?

Normally, in a clean dev-env you should not need this as this dir will be deleted in normal execution flow.

What happened in this triage case is that due to the failure of the Islet test-machinery, this dir used to be left behind. So, upon a re-run of this script, it would fail complaining that the dir already exists.

There possibly is a better place to "fix" this, as a cleanup of the exception 'handling' caught by this bash script. But figuring that out, I felt, is a bit too much work.

Hence, I applied this "Patch-fix" to cope with a dir already existing. If you think using a left-over dir could be problematic, one thing we could do in the `else` is to `rm -rf "${tmp_islet_dir}/*` to clean-up stale contents.

I am not hung-up on this too much, but some fix to allow for a tmp-dir to be around may improve robustness of this script, overall. 

Just a thought ... figure out what seems to make best sense, and do something reasonable.

sample_apps/run_example.sh
```diff
@@ -1108,7 +1108,7 @@ function get_measurement_of_trusted_simple_app_under_islet() {
     # get_measurement_of_app_by_name "islet_example_app.exe"
 
     run_cmd
-    local mrenclave="6190EB90B293886C172EC644DAFB7E33EE2CEA6541ABE15300D96380DF525BF9"
+    local mrenclave="580bd77074f789f34841ea9920579ff29a59b9452b606f73811132b31c689da9"
```
gapisback, line 1109: Thanks for figuring this out ... This hash probably was one of the core causes for the failure.

Qs. How often do you think this kind of hash will change?

I hard-coded it in this script (way long ago) to just get something going.

If we do need to make this more robust, to avoid a hard-coded string, we will need to figure out someway to automatically 'generate' this hash-ID and `grep` for it here, from some other source.

If you think all that hoopla is needed, then feel free to log a different issue, recording the potential problem, and describe a potential solution. One of us (two) could look into improving the situation in future.

On the other hand, if you think this kind of hard-coding is acceptable and it won't change too much, then we can live with it.

sample_apps/run_example.sh
```diff
@@ -1108,7 +1108,7 @@ function get_measurement_of_trusted_simple_app_under_islet() {
     # get_measurement_of_app_by_name "islet_example_app.exe"
 
     run_cmd
-    local mrenclave="6190EB90B293886C172EC644DAFB7E33EE2CEA6541ABE15300D96380DF525BF9"
```
gapisback, line 1109: Suggestion: Given that we have come across an instance where this hash-ID had to be changed (for whatever reasons), can you please add a comment before L1111 documenting the reason why this changed and why it had to be updated.

If you can add a comment capturing the reasons why this changed in this instance, and put down some notes alerting future readers that this kind of [recurring] update may be needed in the future, it will help us guide to this area while troubleshooting future such failures.

The main message would be to document "how-to" figure out the new hash-ID string. (I, myself, would like to know how you figured out how-to get this new hash-ID string.)

third_party/islet/setup.sh
```diff
@@ -29,14 +29,15 @@ ISLET_INC="$HERE/include"
 ISLET_LIB="$HERE/lib"
 
 TARGET_HDR="$ISLET_SDK/include/islet.h"
-TARGET_LIB="$ISLET/out/x86_64-unknown-linux-gnu/debug/libislet_sdk.so"
+TARGET_LIB="$ISLET/out/x86_64-unknown-linux-gnu/debug/libislet_sdk.a"
```
gapisback, line 32: This change from `.so` to `.a` seems to be like build-artifact issue... 

So, I'm not going to worry about this too much, expecting that as this works in CI, things are ok.

gapisback (approved):
Thanks for getting this clean-up done, @bitboom .

Summary seems to be that the failures are **not related** to Ubuntu latest v22.xx upgrade but something else related to build / infra-config around Islet.

For future reference, I suggest you please explain the source of the failure and the fix, and point out in your commit message this is not an Ubuntu-upgrade related issue.

sample_apps/run_example.sh
gapisback: Another approach is to delete the dir if it exists, and create a new one. Something like:

```
if [  -d "${tmp_islet_dir}" ]; then
        rmdir -rf "${tmp_islet_dir}"
fi
```

I think this will both (a) cope with the left-over-dir issue I ran into and will (b) avoid re-using stale dir contents.

gapisback:
@bitboom -- I just merged the PR #243 which comments out failing test-cases.

You should now be able to integrate your work to re-enable ISLET-related failing tests. FYI.

jinbpark:
Nice work, @bitboom !

sample_apps/run_example.sh
bitboom: I restored it :)

sample_apps/run_example.sh
bitboom: I updated comments. Thanks for the suggestion!

PR #317, opened 2026-06-12
Portability fixes for vtpm manufacture and startup scripts

paulengland:
Changed /jlm to ~/$HOME and swapped vtpm-proxy with a socket/socat solution for creating /dev/tpmrm1


jlmucb (commented):
OK

PR #321, opened 2026-06-22
Tpm support

jlmucb:
tpm support, bug fixes better shell fiels


paulengland (approved):
Approve to merge into main

jlmucb:
This upgrade includes bug fixes (including security related ones), better instructions, better scripts, complex measurement support, further support for VM-scale protection (including a keyserver and utilities), some cloud platform support, and most importantly, the Certifier now has TPM support. 

TPM support is important for a few reasons:
    1. It greatly increases the number of possible platforms, both client and
       server.
    2. Cloud vendors, like Amazon, use CC to start foundation software but only
       allow granular support for attestation, measurement and sealing through a
       virtual TPM in the CC attested foundation.  This is the only practical
       way to get CC support on some vendors, and, it extends "real" CC
       protection to a lot of cloud platforms.
    3. If a program is "just" protected by the TPM, absent additional capability
       in the HW (like Nitro or SEV), it does not provide all the CC guarentees
       but it's a lot better than nothing and upgrading to "real" CC
       with the Certifier is easy now.

PR #322, opened 2026-06-27
Tpm2 fixup1

jlmucb:
Tests for cf_utility and simple_app for tpm


paulengland (approved):
Passes all my tests.  Some lingering build warnings in tpm2_support.cc

PR #323, opened 2026-06-30
Tpm2 support fixup2

jlmucb:
More TPM tests


jlmucb:
more tests ans added arg to first pass

